PR #55, opened 2019-05-27
Misc changes + version 0.4.2 bump

rmorshea:
# Summary

Fixes several things:

1. `html_to_vdom` did not return the whole HTML model.
2. Reworks how the transform function is applied to the model after its been converted to VDOM.
3. Fixes typing issue with `Element` function signature.

## Return whole HTML model

If your html looked like this:

```
<div/>
<p/>
```

The function only returned the model for the first `div`. Now it returns a wrapper `div` around both elements.

## Reworks how transform is applied

The transform was a little difficult to work with given the time at which it was applied. It's more intuitive for the transform to get applied from the root node to leafs rather than the other way around. It also makes the underlying code a little easier to reason about.

## Fix `Element` function protocol type

The `Element` functions signature was previously defined by a `Protocol` with a `__call__` attribute, however this didn't really work as expected.

I wanted to enforce `(self: Element, *args: Any, **kwargs: Any)` where `*args, **kwargs` was actually optional (i.e. functions conforming to the protocol could accept any, or no other arguments besides `self`).

See https://github.com/python/mypy/issues/5876 for details.

### DIFF
--- a/requirements/docs.txt
+++ b/requirements/docs.txt
@@ -1,6 +1,6 @@
 # copied from prod.txt
-sanic>=18.12, <19.0
-typing-extensions==3.7.2, <4.0
+sanic >=18.12, <19.0
+typing-extensions >=3.7.2, <4.0
 
 # copied from dev.txt
 pytest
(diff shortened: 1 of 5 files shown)


rmorshea:
Wrong branch

Issue #63, opened 2019-06-10
Remove Asyncio


rmorshea:
While the `asyncio` API looks nice and shiny to me, I've heard some feedback, that given the audience `idom` might serve, the `async/await` syntax might scare them off. Additionally, expensive backend processes block renders of minor changes. For example, the matplotlib example blocks updates to the sliders when the graph image is being created this causes the moving sliders to jitter each time the graph updates.

rmorshea:
`idom` is too deeply invested in `asyncio` so I don't think it can be removed.

Also: #68 resolves problems with blocking so the only reason to this is if people really are averse to `async/await`.

PR #68, opened 2019-06-16
Elements can render in thread executor

rmorshea:
Adds a 'run_in_executor' keyword argument to the `element` decorator that allows users to run a render function in a thread executor. This is useful because certain long running operations will prevent the next render from taking place and thus halt the view until the operation completes. Processing these long running updates in a thread executor unblocks the rendering loop.

In order to make this happen some underlying changes to the `Layout` and communication protocol were made. In short, renders are not run at the time `Layout.render` is called. They are scheduled to be computed immediately and added to a queue when they complete. The `Layout.render` function then pull from that queue. The communication protocol was changed to include the ID of the layout's root element because it may not be the first element to get rendered and thus be placed in the queue.

Other minor changes were made - see `RenderBundle` type to see.

Closes: #63


rmorshea:
There's still a bit of work to do here.

- Tests not passing due to animation timing
- `asyncio.ensure_future` suppresses error messages. Is there an nice way to await those futures?
  - in the `Layout.render` function using `asyncio.wait` instead of the `Queue` might allow them to be awaited.
  - for animation hook there might be a way to do it on unmount?

PR #89, opened 2019-07-25
Shared server update queue put was un-awaited

rmorshea:
Also adds examples for shared server to docs and intro notebook.


src/py/idom/core/render.py
```diff
@@ -75,7 +73,7 @@ def __init__(self, layout: Layout) -> None:
                 del self._models[old_id]
             # append updates to all other contexts
             for queue in self._updates.values():
-                queue.put((src, new, old))
+                await queue.put((src, new, old))
```
rmorshea: This is what caused the bug.

Issue #112, opened 2019-09-05
Change Basic Node Constructor Interface


rmorshea:
`idom.node` allows you to pass children via `*args` and attributes via `**kwargs`. I think it would be cleaner to use if the interface was:

```python
idom.node([...children], {...attributes})
```

Which makes it easier to specify attributes which are Python keywords (e.g. `class` which is proxied by `cls` now).

PR #117, opened 2019-09-15
Rework the standard node constructor

rmorshea:
# Summary

Closes: #112

# Usage

Instead of the following:

```
idom.html.input("click me!", onChange=handler, cls="user-input")
```

We would do this:

```
idom.html.input({"onChange": handler, "class": "user-input"}, ["click me!"])
```

Note the difference between "class" and "cls" as attribute keys - we have to use "cls" because "class" is a keyword. Thus, under the hood their's some ad hoc logic to convert "cls" to "class". To work around this users could do:

```
idom.html.input(..., **{..., "class": "user-input"})
```

But it seems better to have just one way to do things. The same would be true of any attribute names with a "-" in them. Further this encourages a distinction between cameCase being used in models and snake_case being used for Python stuff.


rmorshea:
Need a way to enforce attributes being first:

```
idom.html.img({"onClick": my_handler}, [children, ...])
```

Instead of:

```
idom.html.img([children, ...], {"onClick": my_handler})
```

Having consistent order will just make things more readable.

PR #120, opened 2019-09-26
Transpile HTML Template Syntax

rmorshea:
Add an optional extension to the Python language for writing HTML with an [fstring](https://www.python.org/dev/peps/pep-0498/)-like template syntax:

```python
# coding=html
import idom

size = "30px"
text = "Hello!"

model = html"""
<div height={size} width={size} >
    <p>{text}</p>
</div>
"""

assert model == {
    "tagName": "div",
    "attributes": {"height": "30px", "width": "30px"},
    "children": ["\n    ", {"tagName": "p", "children": ["Hello!"]}, "\n"],
}
```

Can be used by import idom at the application endpoint before importing modules that use the syntax, or permanently install the extension so you can use it anywhere with `# coding=html` using:

```
htm-pyx register
```


src/py/idom/codec/transpile.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def handle_endtag(self, tag: str) -> None:
+        self._write("]), ")
+
+    def handle_startendtag(self, tag: str, attrs: Sequence[Tuple[str, str]]) -> None:
+        self.handle_starttag(tag, attrs)
+        self.handle_endtag(tag)
+
+    def parse_starttag(self, i: int) -> int:
+        self.__starttag_text = None
```
rmorshea: Add note that this was copy-pasted from:
https://github.com/python/cpython/blob/master/Lib/html/parser.py

src/py/idom/codec/transpile.py
```diff
@@ -0,0 +1,161 @@
...
+        return endpos  # type: ignore
+
+    def _write(self, text: str) -> None:
+        self._cursor += self._code.write(text)
+        self._code.seek(self._cursor)
+
+    def _load_exprs(self, text: str) -> str:
+        parts: List[str] = []
+        for is_expr, string in split_fstr_style_exprs(text):
+            if string:
```
rmorshea: Just don't return empty strings from `split_fstr_style_exprs`

rmorshea:
The transpiler matches on strings of the form:

```
"...html"
```

It should make sure that `html"` is not the end of a string.

rmorshea:
Transpiler moved to [htm-pyx](https://github.com/rmorshea/htm-pyx).

Just need to move docs to htm-pyx and reference them from IDOM.

rmorshea:
Superseded by #121

PR #155, opened 2020-04-24
fixes #154 views do not redisplay on refresh in Jupyter Notebook

### LINKED ISSUE #154: Views Disapear After Refreshing Jupyter Notebook
This was actually intentional, but should really be a temporary solutions since it's rather inconvenient. In short we did this because requesting a resource from a different origin must be successful otherwise future requests to that resource will be blocked by the browser.

So if you restarted your kernel and refreshed the notebook page your IDOM widget views would fail to display when you re-ran it because the pre-existing view attempted to connect to the IDOM webserver when it didn't exist yet.

To work around this we temporarily displayed a script that set `document.idomServerExists = true` so that views could check that flag before they displayed. This will be changed soon though.

rmorshea:
make a `fetch` request with `{mode: "no-cors"}` to check if the IDOM webserver is running before requesting resources.

### DIFF
--- a/idom/widgets/display.py
+++ b/idom/widgets/display.py
@@ -2,7 +2,6 @@ import uuid
 import json
 from typing import Any
 from urllib.parse import urlparse
-from IPython import display as _ipy_display
 
 
 def display(kind: str, *args: Any, **kwargs: Any) -> Any:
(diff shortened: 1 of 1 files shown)


rmorshea:
Got feedback that this fixed the issue.

PR #156, opened 2020-04-28
Fix #157 add per-client view parameters via HTML query

### LINKED ISSUE #157: Allow Multiple Active Views in Jupyter Notebook
This limitation is due to the fact that there was no way to provide per-client parameters. Enabling clients to pass parameters to IDOM elements upon connection would enable them to display a unique view to each client based on which parameters were passed. There are probably other use cases for per-client view parameters beside just in the Jupyter Notebook so this is definitely worth doing.

rmorshea:
Along with these per-client view parameters we add a `multiview` widget that allows many elements to be dynamically and imperatively added to an already mounted view.


rmorshea:
Still need to update tests, intro notebook, and docs...

PR #172, opened 2020-05-15
use setuptools_scm version

rmorshea:


rmorshea:
This only failed because I accidentally had the wrong event type attached to the GH action.

PR #215, opened 2020-09-11
use preact instead of react

rmorshea:
Preact is much smaller and thus loads more quickly.


rmorshea:
nevermind... not worth it

PR #320, opened 2021-03-07
Upgrade Snowpack

rmorshea:
We're stuck on the last major version. We should try to catch up before it becomes too painful to upgrade.

As part of this upgrade we're adding a new `idom.config` module in which there a options that can be modified by downstream applications to meet their specific needs. These options can be changed using environment variables, and a subset can be changed at runtime. For example, there is now an `idom.config.IDOM_CLIENT_BUILD_DIR` option that can be `set()` to change what directory IDOM uses to store its client.

cc: @philippjfr - these changes will impact [Panel's usage of IDOM](https://github.com/holoviz/panel/pull/2004/files#diff-449a0429f825e4e687317b9985ca1abb6f36a992047279dde1e486afea5fc032R146-R148). The addition of `idom.config` should make this use case less brittle in the future though.


philippjfr:
Thanks for the heads up and for refactoring the config. That will indeed be a lot cleaner.

Issue #298, opened 2021-02-25
IDOM causes race condition in AnyIO


rmorshea:
See this issue for more info and potential solutions: https://github.com/agronholm/anyio/issues/210

Issue #352, opened 2021-04-22
SharedStateServer doesn't send update to client after refresh


brentbaum:
I have a simple app with a few buttons. When the user clicks on a button, the text next to the button toggles between "asserted=False" and "asserted=True".

<img width="388" alt="Screen Shot 2021-04-22 at 8 51 47 AM" src="https://user-images.githubusercontent.com/868955/115746199-22267400-a362-11eb-83b2-52f3245fd68b.png">

The first time a user brings up the app everything works as expected. The network tab looks like this.

<img width="525" alt="Screen Shot 2021-04-22 at 8 51 51 AM" src="https://user-images.githubusercontent.com/868955/115746669-821d1a80-a362-11eb-93d0-833382b44ecc.png">

After a refresh the app keeps sending events but stops receiving them.

<img width="546" alt="Screen Shot 2021-04-22 at 8 52 07 AM" src="https://user-images.githubusercontent.com/868955/115746706-8a755580-a362-11eb-8a3b-0a9b92a19e24.png">

This behavior is consistent across any components that depend on state. Notably, the user is unable to change the value of inputs.

Server code:

```
def start_app(dispatch):
    from idom.server.sanic import SharedClientStateServer

    # we use a redux-like dispatch structure to send events to outside the idom event, then update state
    # with a callback stored outside the app. 
    # I've tried removing this particular piece of functionality and the problem persists.
    root = lambda: Panel(dispatch)
    app = SharedClientStateServer(root)
    return app.daemon("0.0.0.0", 3040)
```

Any ideas on the root cause or a way to mitigate? The PerClientStateServer throws a CannotAccessError due to our rapid state updates, logging `(f"Did not update {component} - resources of {self} are closed")` so it's a no-go (sorry, I've been waiting to file an issue on that one!)

Thanks!

rmorshea:
Did this issue arise after the latest update?

brentbaum:
@rmorshea I've had this issue since 0.22 and haven't tested before that. 

You got it, the `CannotAccessError` (really, just the log message but that's triggered by a `CannotAccessError`) happens after refresh.

rmorshea:
Ok, it looks like I'm able to reproduce, so I'll look into that.

[...]

PR #353, opened 2021-04-24
Refactor Dispatchers

rmorshea:
Close: #298, #352 

This switches from a class-based approach to one that leverages pure functions. While the logic itself isn't any simpler, it was easier to implement (and hopefully understand) it correctly. This conversion was motivated by several bugs that had cropped up and that wouldn't be easily fixable without substantial changes refactor or not.

The one thing that still gets me is the lack of uniformity in usage, but I suppose that's alright given how different the "shared" vs "single" view implementations are.


Archmonger:
Hell yeah! Whenever this build gets moved over to PyPi I will begin Conreq integrations.

Issue #253, opened 2020-10-29
Allow customization of built-in IDOM client


rmorshea:
It would be great if it were possible to do some basic templating on the IDOM client web page. Without this, if you want to include custom styles or JS in the `<head />` of the HTML markup you have to completely re-create the client.

The alternative I suppose would be to make it easier to create custom clients. I'm not entirely sure what that would look like though. Maybe just a cookie cutter repo? The downside there of course is that you can't automatically update the repo if you update the template.

rmorshea:
This came up again: https://github.com/idom-team/idom/discussions/400

rmorshea:
@brentbaum The least brittle option I've been able to think of is to try and load a `/custom.js` script in `index.html`:

```html
<script>
  import("/custom.js")
    .then(() => console.log("Loaded custom.js script"))
    .catch(() => console.log("No custom.js script found"));
</script>
```

This would allow you to serve a `custom.js` file that could manipulate the page in any way you might want:

```python
from tempfile import NamedTemporaryFile

from sanic import Sanic

import idom
from idom.server.sanic import PerClientStateServer

temp = NamedTemporaryFile()
open(temp.name, "w+").write("document.title = 'Hello World';")  # set the page title on load

app = Sanic()
app.static("/custom.js", temp.name, name="custom_js", content_type="text/javascript")

HelloWorld = idom.component(lambda: idom.html.h1("Hello World"))
PerClientStateServer(HelloWorld, app=app).run("127.0.0.1", 8000)
```

Thoughts on this approach?

rmorshea:
I suppose some of these details could be abstracted away by making this a `server_config` option:

```python
import idom

idom.run(MyComponent, server_config={"custom_js": "path/to/custom.js"})
```

brentbaum:
I like the second approach quite a bit! The killer here is that React doesn't allow you to render and execute <script> tags by default - I tried for an hour yesterday to get a simple `<script>alert("hi!")</script>` example to run before remembering. If I could run js inline the same way I can define inline css this wouldn't be a problem.

rmorshea:
@brentbaum what do you think about making something like this into a component that you'd just include in your layout?

```python
from idom import Javascript

my_javascript = Javascript("path/to/my/script.js")

@idom.component
def MyApp():
    return idom.html.div(my_script, ...)
```

PR #436, opened 2021-07-17
Enable standard JS scripts to be linked in as a module

rmorshea:
Fixes: #253

This accomplished by mildly abusing the custom component system.
In short, we create a no-op module interface so IDOM will load
the file, but won't actually do anything afterwards


rmorshea:
There's probably a better way to implement this.

PR #517, opened 2021-10-17
Render components in a ThreadPool

Archmonger:
Uses a ThreadPoolExecutor to execute sync views in an asynchronous fashion.
This ensures IO bound components do not permanently lock the GIL.

### DIFF
--- a/src/idom/core/layout.py
+++ b/src/idom/core/layout.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 import abc
 import asyncio
 from collections import Counter
+from concurrent.futures import ThreadPoolExecutor
 from functools import wraps
 from logging import getLogger
 from typing import (
@@ -35,6 +36,8 @@ from .vdom import validate_vdom_json
 
 
 logger = getLogger(__name__)
+
+RENDER_POOL = ThreadPoolExecutor()
 
 
 class LayoutUpdate(NamedTuple):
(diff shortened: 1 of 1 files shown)


rmorshea:
So unfortunately, this change doesn't parallelize things in the way you might imagine. The main reason is that calls to `Layout.render` need to happen serially. That is the order in which things render might matter. We can see [here](https://github.com/idom-team/idom/blob/main/src/idom/core/dispatcher.py#L212-L216) that we do one render after the other.

The other thing is that there's already a solution to IO bound operations, and that is the `use_effect` hook. Let's say you have a component which does some long IO bound operation. Well, the way to achieve that is to put this long operation inside an "effect":

```python
import idom

@idom.component
def WaitOnResource():
    resource, set_resource = idom.hooks.use_state(None)

    @idom.hooks.use_effect
    async def await_resource():
        result = await long_io_operation()
        set_resource(result)

    if resource is None:
        return make_fallback_display()
    else:
        return make_actual_display(resource)
```

The way this works is that the first time this component renders the `resource` will be `None` and we display a fallback to the user (maybe a spinner of some sort). However, we also registered a callback with `use_effect`. The `use_effect` hook tells the layout to schedule the given callback to run asynchronously after its current render completes. Thus, we can wait on our long running IO operation without blocking the layout from rendering in the meantime. Once that operation completes, we get back a `result` and `set_resource` causing the component to re-render. This time though, we'll `make_actual_display` since the `resource` is no longer `None`.

PR #518, opened 2021-10-19
add class based component implementation

rmorshea:
This pull request is provisional, and it may never be merged. Hopefully it can serve as a useful reference if this is asked about in the future.

# What This Does

This PR allows you to define components using classes similarly to how you would in React. Python's `__setattr__` magic method makes it a bit easier to implement though since you no longer need an `update` method. Instead the instance attributes can be assigned directly.

```python
@idom.component
class Counter:
    def __init__(self, count: int = 0):
        self.initial_count = self.count = count

    def render(self):
        return idom.html.div(
            f"Count: {self.count}",
            idom.html.button({"onClick": self.reset}, "Reset"),
            idom.html.button({"onClick": self.increment}, "+"),
            idom.html.button({"onClick": self.decrement}, "-"),
        )

    def increment(self, event):
        self.count += 1

    def decrement(self, event):
        self.count -= 1

    def reset(self, event):
        self.count = self.initial_count
```

# Limitations

This PR does not allow class-based components to incur side effects. We may leave that as an intentional limitation though in order to try and push people towards functional components because they encourage better patterns of development.

# Disadvantages

1. Having two ways to do the same thing might be confusing. Do we introduce class-based or function-based components first?
2. Class-based components are less composable and can encourage anti-patterns - most new React libraries use hooks for this reason.
3. It will likely be strange for users to see that mutating attributes of their components will not trigger updates. For example, appending to a list won't trigger a re-render.
4. It probably isn't that big of a deal, but I suspect these class-based components might be less performant - users would have to understand that every attribute assignment has side effects they may not be aware of.

# Advantages

- If used sparingly, class-based components could enable some useful imperative interfaces. This would probably be a bad use case, but in the `Counter` example above a parent component could increment the count by mutating its attributes.
- Class based components could serve as a useful introduction to the execution model. The idea being that we avoid scaring off new users, but intentionally limit class-based capabilities in order to push them towards learning/using function-based ones.

### DIFF
--- a/docs/source/examples/snake_game_class.py
+++ b/docs/source/examples/snake_game_class.py
@@ -1,0 +1,194 @@
+import asyncio
+import enum
+import random
+import time
+from dataclasses import dataclass, replace
+
+import idom
+
+
+class GameState(enum.Enum):
+    init = 0
+    lost = 1
+    won = 2
+    play = 3
+
+
+@idom.component
+class GameView:
+
+    state = GameState.init
+
+    def render(self):
+        if self.state == GameState.play:
+            return GameLoop(self, grid_size=6, block_scale=50)
+
+        start_button = idom.html.button(
+            {"onClick": lambda event: idom.set_state(self, GameState.play)},
+            "Start",
+        )
+
+        if self.state == GameState.won:
+            menu = idom.html.div(idom.html.h3("You won!"), start_button)
+        elif self.state == GameState.lost:
+            menu = idom.html.div(idom.html.h3("You lost"), start_button)
+        else:
+            menu = idom.html.div(idom.html.h3("Click to play"), start_button)
+
+        menu_style = idom.html.style(
+            """
+            .snake-game-menu h3 {
+                margin-top: 0px !important;
+            }
+            """
+        )
+
+        return idom.html.div({"className": "snake-game-menu"}, menu_style, menu)
+
+
+class Direction(enum.Enum):
+    ArrowUp = (0, -1)
+    ArrowLeft = (-1, 0)
+    ArrowDown = (0, 1)
+    ArrowRight = (1, 0)
+
+
+@dataclass(frozen=True)
+class GameLoopState:
+    direction: tuple[int, int]
+    snake: list[tuple[int, int]]
+    food: tuple[int, int]
+
+
+@idom.component
+class GameLoop:
+    def __init__(self, game_view, grid_size, block_scale):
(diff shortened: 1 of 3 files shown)


Archmonger:
On a different note, I definitely think this is vastly more beginner friendly compared to the functional components.

Archmonger:
### `use_effect` Abstraction
In terms of abstracting to the `use_count_down` example, couldn't that be automated through wrapping the `Component` as it passes through the IDOM component registration logic and configuring `use_effect` discreetly?

This is with the assumption that the limitation is that `use_effect` isn't getting called in the CBC example. If not then I'd need some more context.

---

### Separating Methods within the Class
I personally think this is a benefit. Increases readability and allows for re-use of the `reset_count` logic within other parts of the component.

I'd actually argue this might be more composable than the functional example. In the CBC example, now `reset_count(self)` can be pushed out of the class if needed and re-used for several different components.

In fact, hooks could also now be made generic if needed. For example, `effect(self)` could be pushed out and re-used, if that situation comes across for more complex examples.

Basically the only thing that would need to remain in every class is `__init__` and `render(self)`, encouraging a design pattern of one component = one html segment.

---

### Encouraged Design Patterns
Although I did mention beginner friendliness, I would also like to mention that the functional example you posted is very difficult to read, also making it difficult to debug.

It only took one glace on the class based example to fully understand what is going on, while I had to re-read and mentally stack trace the functional example roughly three times to follow the same logical flow.

If a simple decrementing counter required me to constantly bounce my eyes up and down the code, it's telling that the functional syntax is not easily digestible.

rmorshea:
After looking into it for a bit, it turns out that making `state` depend on a constructor parameters is an anti pattern. So maybe we can just have people assign it at the class level.

rmorshea:
I'm hoping that you'll be able to use hooks with class-based components. To do that, I'll need to enforce the rules of hooks in the `render()` method and thus I need to be able to statically identify which classes are component classes.

rmorshea:
A few more slight changes:

- split `effect()` into `render_effect()` (runs after every render) and `mount_effect()` (runs once after first render)
- allow `state` attribute to be defined with `@property`

With these changes I was able to rewrite the snake example: https://gist.github.com/rmorshea/84d66b2ef61ae38171c8066bb4368d36

Archmonger:
This PR should probably be shelved and looked at within the v3 release cycle.

rmorshea:
I'm going to close this for now.

PR #547, opened 2021-12-15
improve api section

rmorshea:
The right-hand sidebar now looks like this:
image


Archmonger (approved):
Much improves things.

PR #566, opened 2022-01-05
implement the use_debug_value hook

rmorshea:
part of: https://github.com/idom-team/idom/issues/203

### DIFF
--- a/src/idom/core/hooks.py
+++ b/src/idom/core/hooks.py
@@ -24,7 +24,10 @@ from typing import (
 
 from typing_extensions import Protocol
 
+from idom.config import IDOM_DEBUG_MODE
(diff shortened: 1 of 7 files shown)


src/idom/core/hooks.py
```diff
@@ -392,6 +395,54 @@ def use_ref(initial_value: _StateType) -> Ref[_StateType]:
     return _use_const(lambda: Ref(initial_value))
 
 
+if IDOM_DEBUG_MODE.current:
```
Archmonger, line 398: Would be cleaner to use a guard statement directly within `use_debug_value`

```python
def use_debug_value( ... ) 
   if IDOM_DEBUG_MODE.current:
      return
   ...
```

Issue #203, opened 2020-09-01
Not All React's Hooks Are Implemented


rmorshea:
Not all of [React's builtin](https://reactjs.org/docs/hooks-reference.html) hooks have been implemented. If you have a pressing need for a missing hook leave a comment here.

- [x] `use_state`
- [x] `use_effect`
- [x] `use_context`
- [x] `use_reducer`
- [x] `use_callback`
- [x] `use_memo`
- [x] `use_ref`
- [x] `use_debug_value`

Archmonger:
@rmorshea Which hooks are still under development?

rmorshea:
I think the only one not implemented at this point is `useContext`.

rmorshea:
This is done now!

PR #585, opened 2022-01-14
Add Use Context Hook

rmorshea:
Task from: #203


Archmonger:
Should this PR encompass https://github.com/idom-team/idom/discussions/594 or should I open a separate issue specifically for per-server `RequestContext`?

rmorshea:
I think it should be a separate issue. `use_context` has other applications besides just the request context

Issue #584, opened 2022-01-14
Fix unhandled case in module_from_template


rmorshea:
### Current Situation

Currently we do not handle the case where the target-file already exists in `module_from_template`:

https://github.com/idom-team/idom/blob/0bf58771ac9fe0fa04853fbc33a2fdf614aaaed3/src/idom/web/module.py#L144-L150

### Proposed Actions

Add a `replace_existing` option and raise an error if it is False and the target file exists.

PR #586, opened 2022-01-14
Log on web module replacement instead of error

rmorshea:
Closes: #584

### DIFF
--- a/src/idom/web/module.py
+++ b/src/idom/web/module.py
@@ -1,10 +1,12 @@
 from __future__ import annotations
 
+import filecmp
 import shutil
-from dataclasses import dataclass
+from dataclasses import dataclass, replace
 from functools import partial
 from pathlib import Path
 from string import Template
+from tempfile import NamedTemporaryFile
 from typing import Any, List, NewType, Optional, Set, Tuple, Union, overload
 from urllib.parse import urlparse
 
@@ -74,6 +76,9 @@ def module_from_url(
     )
 
 
+_FROM_TEMPLATE_DIR = "__from_template__"
+
+
 def module_from_template(
     template: str,
     package: str,
@@ -83,6 +88,7 @@ def module_from_template(
     resolve_exports: bool = IDOM_DEBUG_MODE.current,
     resolve_exports_depth: int = 5,
     unmount_before_update: bool = False,
+    replace_existing: bool = False,
 ) -> WebModule:
     """Create a :class:`WebModule` from a framework template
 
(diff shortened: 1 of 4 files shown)


Archmonger:
From a Django caching perspective, it would be ideal if it only replaced if the file contents are not identical.

rmorshea:
That is what's implemented here. If the file contents differ and `replace_existing` is `False` you get an error instead of replacing the files.

src/idom/web/module.py
```diff
@@ -83,6 +88,7 @@ def module_from_template(
     resolve_exports: bool = IDOM_DEBUG_MODE.current,
     resolve_exports_depth: int = 5,
     unmount_before_update: bool = False,
+    replace_existing: bool = False,
```
Archmonger, line 91: A better default for this is `True`. Most developers would expect this behavior.

src/idom/web/module.py
```diff
@@ -83,6 +88,7 @@ def module_from_template(
     resolve_exports: bool = IDOM_DEBUG_MODE.current,
     resolve_exports_depth: int = 5,
     unmount_before_update: bool = False,
```
Archmonger, line 90: Would be nice to have control over `symlink=...` up to at this level.

src/idom/web/module.py
rmorshea: I don't want someone to silently and accidentally replace a module they didn't mean to. Because this directory is shared globally, libraries need to cooperate with respect to what they name their bundled module files. Ideally every module should correspond to a package published to NPM, but we can't necessarily rely on that.

src/idom/web/module.py
rmorshea: Because the contents of these modules are generated from a template there isn't a file to symlink to.

src/idom/web/module.py
Archmonger: Perhaps a global setting that can configure this value then?

At least within Django-IDOM, it'd be nice to have some way to set this as default.

rmorshea: > I'm thinking to put the files directly into the Django-IDOM app directory

So copying files from `IDOM_WEB_MODULES_DIR` into the app directory, or the other way around?

Archmonger: My thought process was to either set `IDOM_WEB_MODULES_DIR` to `<django_idom>/data/web_modules/` by default and letting the user further configure this directory through `settings.py`.

The alternatives are symlinking or copying them over.

At some point, we're going to be putting the web module files directly into `django_idom/static/` whenever I get around to [this ticket.](https://github.com/idom-team/django-idom/issues/4)

rmorshea: Perhaps I could log a warning instead of raising?

Archmonger: Agreed, it's not a critical failure so it shouldn't be an exception.

PR #631, opened 2022-01-29
Fix state mismatch during component update

### LINKED ISSUE #629: State mismatch during component update
### Current Situation

Currently, updating any child component state can result in `AssertionError: state mismatch during component update - key '61e47127-9d3f-4954-8235-4d99122134c1'!=4`

This can be previewed within Conreq's app store branch.

### Proposed Actions

Update `layout.py` to not cause this exception.

### Work Items

- [ ] Develop a fix for state mismatch
- [ ] Release 0.35.5 to resolve this

Archmonger:
- fix #629
- Bumps to version 0.35.5

Something worth thinking about is whether this error had actually existed due to the previous state hook not being deleted from memory. Just want to ensure there are no memory leaks occurring.

### DIFF
--- a/src/idom/core/layout.py
+++ b/src/idom/core/layout.py
@@ -216,11 +216,6 @@ class Layout:
             pass
         else:
             key, index = new_state.key, new_state.index
-            if old_state is not None:
-                assert key == old_state.key, (
-                    "state mismatch during component update - "
-                    f"key {key!r}!={old_state.key!r} "
-                )
             parent.children_by_key[key] = new_state
             # need to do insertion in case where old_state is None and we're appending
             parent.model.current["children"][index : index + 1] = [


Archmonger:
`nox -s update_version 0.35.5` doesn't appear to work on my end.

Results in this error:
```cmd
nox > Error while collecting sessions.
nox > Sessions not found: 0.35.5
```

VERSION
```diff
@@ -1 +1 @@
-0.35.4
+0.35.5
```
rmorshea, line 1: Let's hold off here for the moment. Technically this change includes the script tag so, for the sake of simplicity, I think I'm gonna make the suggested changes to the `script` and make a `0.36.0` release.

rmorshea (changes requested):

VERSION
rmorshea: Err. Maybe we should release this but just not announce the `script`.

Archmonger: I'll just pull the version bump out of this PR, and we can create a 0.36.0 release separately.

Archmonger:
Let me know if you think this is safe.

I'm slightly worried this exception may have occurred due to an incomplete wipe of component hooks on render updates. If so, then that would be a memory leak.

But I don't know enough about the layout call stack to say anything with certainty.

rmorshea (approved):

rmorshea:
I think it's ok. I'm pretty certain I introduced this because I didn't really understand how keys worked very well.

rmorshea:
We unmount on the same condition [here](https://github.com/idom-team/idom/blob/c46cf9ee495bfba8fd588b1c8dc40c00e6e2d848/src/idom/core/layout.py#L243)

rmorshea:
I'll have time to release this tomorrow.

rmorshea (approved):

Issue #592, opened 2022-01-16
Enforce Docstring Style Guidelines


rmorshea:
### Current Situation

Right now we don't enforce any particular docstring style

### Proposed Actions

Enforce Google-style docstrings with Pydocstyle.

Pydocstyle hasn't cut a release that includes support for hanging indents, but support has been merged to main. Thus we should install from this commit until a release is made:

```
pip install "git+https://github.com/PyCQA/pydocstyle.git@bd4993345a241bd0d5128f21de56394b1cde3714#egg=pydocstyle"
```

### Work Items

- [ ] enforce docstring style with pydocstyle
- [ ] add comment explaining why we install from a specific commit on github

PR #667, opened 2022-02-11
Fix 663

rmorshea:
closes: #592


Archmonger:
Does this also fix #592?

rmorshea:
Yeah, that looks like I mistakenly branched off some other feature branch. I'll correct that.

Issue #595, opened 2022-01-16
Remove Flake8 From Pre-commit


rmorshea:
### Current Situation

Right now we run Flake8 in our pre-commit hooks. It would be best if we only ran tools that can automatically fix issues in pre-commit hooks. All other issues will be discovered in CI.

### Proposed Actions

Remove Flake8 from pre-commit checks

Issue #538, opened 2021-11-21
Allow NoneType and Iterable returns within components


Archmonger:
### NoneType
There are instances where the user may genuinely want to pretend like a component doesn't exist (ex. Django IDOM conditional rendering based on authentication). Currently, exceptions are generated if a component returns NoneType. 
```bash
2021-11-21T13:15:57-0800 | ERROR | Failed to render homepage(1408449473488, websocket=<django_idom.websocket_consumer.IdomAsyncWebsocketConsumer object at 0x00000147EB0C4B80>)
Traceback (most recent call last):
  File "C:\Users\Username\Repositories\Conreq\.venv\lib\site-packages\idom\core\layout.py", line 201, in _render_component
    self._render_model(old_state, new_state, raw_model)
  File "C:\Users\Username\Repositories\Conreq\.venv\lib\site-packages\idom\core\layout.py", line 236, in _render_model
    new_state.model.current = {"tagName": raw_model["tagName"]}
TypeError: 'NoneType' object is not subscriptable
```

### Iterables
In many situations, not allowing for iterable return type increases the the DOM depth can be unnecessarily. For example, in my circumstances, my components are almost always returning nothing but several other components. I need to add an unnecessary div wrapper to get the component rendering working. From Django IDOM's side, this makes for the top-level component div, and then a second wrapper div directly below it.
![image](https://user-images.githubusercontent.com/16909269/142779382-00a41d49-cd6c-4c6f-a77b-6e1fb1aa9198.png) 
```bash
2021-11-21T13:18:52-0800 | ERROR | Failed to render homepage(1628060950176, websocket=<django_idom.websocket_consumer.IdomAsyncWebsocketConsumer object at 0x0000017B0FF1B6A0>)
Traceback (most recent call last):
  File "C:\Users\Username\Repositories\Conreq\.venv\lib\site-packages\idom\core\layout.py", line 201, in _render_component
    self._render_model(old_state, new_state, raw_model)
  File "C:\Users\Username\Repositories\Conreq\.venv\lib\site-packages\idom\core\layout.py", line 236, in _render_model
    new_state.model.current = {"tagName": raw_model["tagName"]}
TypeError: list indices must be integers or slices, not str
```

### Proposed Changes

Allow for NoneType and Iterable return types in components.

### Implementation Details

Modify `idom.core.layout` to detect these situations and handle them differently than the standard VDOM return type.

rmorshea:
[...]

PR #672, opened 2022-02-16
fix #652

### LINKED ISSUE #652: All child states wiped upon any child mutation
### Current Situation

If two children obtain their keys from a parent element, both children's `use_state` will be wiped upon either child changing it's `key`.

```python
from idom import component, html, run, hooks
from uuid import uuid4


@component
def Parent():
    # Position 0 is the key for child 0
    p_state, set_p_state = hooks.use_state(("0", "1"))

    async def button_click(_):
        print("Button clicked", p_state)
        set_p_state(("0", uuid4().hex))

    return html.div(
        html.button({"onClick": button_click}, "click me"),
        Child0(p_state, set_p_state),
        Child1(p_state, set_p_state),
    )


@component
def Child0(p_state, set_p_state):
    state, set_state = hooks.use_state(0)

    @hooks.use_effect
    async def whatever():
        if state:
            return
        print("Child 0 has no state")
        set_state(1)

    return html.div(
        f"1) state: {state} key: {p_state[0]}",
        key=p_state[0],
    )


@component
def Child1(p_state, set_p_state):
    state, set_state = hooks.use_state(0)

    @hooks.use_effect
    async def whatever():
        if state:
            return
        print("Child 1 has no state")
        set_state(1)

    return html.div(
        f"1) state: {state} key: {p_state[1]}",
        key=p_state[1],
    )


run(Parent)

````

### Proposed Actions

Investigate and resolve this parent-child key abnormality

### Work Items

See above

rmorshea:
closes: #652
closes: #538
closes: #595 

In short, because components do not have a node in the model state tree
managed by the layout, keys for elements in the components were being
registered to the parent of the component rather than to a node for
the component which contained the elements.

The short-term solution was to make it so components wrap their
contents in a div without a key. The long-term solution is to
refactor the Layout() and give components a dedicated node in the
model state tree


rmorshea:
cc: @Archmonger

Archmonger:
Two in one special 😄

PR #696, opened 2022-03-02
actually fix #684

### LINKED ISSUE #684: Input's onChange Event Handler Misses Key Strokes?
### Discussed in https://github.com/idom-team/idom/discussions/668

<sup>Originally posted by **jkrobicki** February 14, 2022</sup>
Description
When I fill in the input fast it reacts slowly and miss some letters. In the example I try to write "write here the name" .
name iput
</div>

rmorshea:
closes: #684 

Because we handle events asynchronously, we must leave the value
uncontrolled in order to allow all changed committed by the user to be
recorded in the order they occur. If we don't, the user may commit
multiple changes before we render resulting in prior changes being
overwritten by subsequent ones. Instead of controlling the value, we set
it in an effect. This feels a bit hacky, but I can't think of a better
way to work around this problem.

### DIFF
--- a/docs/source/_custom_js/package-lock.json
+++ b/docs/source/_custom_js/package-lock.json
@@ -19,7 +19,7 @@
       }
     },
     "../../../src/client/packages/idom-client-react": {
-      "version": "0.37.0",
+      "version": "0.37.1-a1",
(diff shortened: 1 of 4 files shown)


rmorshea:
This is only a partial solution though. If the server modifies the value of the input, then we're basically back to the same issue as before. I'm not really sure how to solve that unfortunately...

Issue #573, opened 2022-01-10
Serialize Form Data


rmorshea:
### Current Situation

Currently, `onSubmit` events for `form` elements do not serialize the data contained within them. To do this we need to serialize the `form.elements` attribute. 

### Proposed Changes

```
diff --git a/src/client/packages/idom-client-react/src/event-to-object.js b/src/client/packages/idom-client-react/src/event-to-object.js
index b11e653..80b098b 100644
--- a/src/client/packages/idom-client-react/src/event-to-object.js
+++ b/src/client/packages/idom-client-react/src/event-to-object.js
@@ -49,6 +49,13 @@ const elementTransformCategories = {
       return {};
     }
   },
+  hasElements: (element) => {
+    const result = {};
+    for (key in element.elements) {
+      result[key] = serializeDomElement(element.elements[key]);
+    }
+    return result;
+  }
 };
 
 function defaultElementTransform(element) {
@@ -69,6 +76,7 @@ const elementTagCategories = {
   ],
   hasCurrentTime: ["AUDIO", "VIDEO"],
   hasFiles: ["INPUT"],
+  hasElements: ["FORM"],
 };
 
 const elementTransforms = {};
```

Archmonger:
There's going to need to be considerations made for file fields. We're also going to need to make [considerations related to backpressure](https://web.dev/websocketstream/), especially since there's the potential to upload files with an uncapped size.

Probably worth generalizing the data serialization (perhaps even as a separate package?). If it's generic enough, there's a possibility that down the line it would allow for using the WS to stream video.

rmorshea:
I think there's an issue (or discussion) about serializing files.

Could you make file serialization into an issue if it isn't one and create an separate issue for dealing with back-pressure?

rmorshea:
@acivitillo this seems like a less-trivial first issue.

Archmonger:
File serialization issue appears to exist: https://github.com/idom-team/idom/issues/574

Shouldn't backpressure be handled within the file serialization issue? Feels strange creating a new issue to handle an _not yet developed_ limitation.

rmorshea:
I think back-pressure is a general problem. Any event, not just those with files, could take a long time for the network to send and the server to process which would thus cause things to back up.

Archmonger:
Something else I'm generally thinking about on this PR is there's going to need to be some degree of consideration for Django here.

[...]

PR #699, opened 2022-03-03
Form serialize

acivitillo:
closes: #573


rmorshea:
I'll have time to take a look at this over the weekend

acivitillo:
yes, do note it's broken, I send you the diff to understand how you want to approach `form` serialization as `form` contains multiple elements

rmorshea:
@acivitillo, I've fixed up the serialization implementation. There were a couple things that were wrong about my suggestions on how to do this.  

Could you address the tests? In addition to the `elements` attribute, I've serialized the `name` attribute too.

Archmonger:
Serializing form attributes into the URL would be a nice feature if we could get it working cleanly. Might help with form value persistence upon WS disconnect?

We should do some investigation later down the line to figure out what React apps typically do for forms.

rmorshea:
To properly handle form URLs we'll need [a wrapper](https://github.com/idom-team/idom-router/issues/7) around the standard form element.

Archmonger:
What do we plan on doing about file fields

rmorshea:
Not entirely sure. We can address separately to address https://github.com/idom-team/idom/issues/574.

Archmonger (dismissed):

PR #700, opened 2022-03-05
buffer onChange events by 100ms

rmorshea:
I think this might help reduce flicker observed in Brave browser with fast inputs. It should also improve performance overall by reducing the number of events being sent.


rmorshea:
never mind, flicker in brave is intermittent.

Issue #678, opened 2022-02-18
Update IDOM Sanic Requirement


rmorshea:
It turns out that IDOM requires Sanic>19,<19.12. This is a slim range of versions because of [this issue](https://github.com/idom-team/idom/issues/591)

### Discussed in https://github.com/idom-team/idom/discussions/674

<div type='discussions-op-text'>

<sup>Originally posted by **gebrian0** February 16, 2022</sup>
```
>>> import idom
>>> from idom import component, html, run
>>> 
>>> 
>>> @component
... def App():
...     return html.h1("Hello, ddWorld!")
... 
>>> 
>>> run(App)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.9/site-packages/idom/server/prefab.py", line 58, in run
    server = server_type(component, server_config, app)
  File "/usr/local/lib/python3.9/site-packages/idom/server/sanic.py", line 74, in PerClientStateServer
    return SanicServer(app)
  File "/usr/local/lib/python3.9/site-packages/idom/server/sanic.py", line 111, in __init__
    app.register_listener(self._server_did_start, "after_server_start")
AttributeError: 'Sanic' object has no attribute 'register_listener'
```
</div>

Issue #657, opened 2022-02-08
Reconsider how we use/advertise idom.run()


Archmonger:
### Current Situation

`idom.run()` probably shouldn't be the primary advertised way of using IDOM. Especially in it's current state.

Using a development webserver for production use is a big no-no, and by prominently displaying `idom.run()` as the first steps to using IDOM shows we are encouraging people to do just that. I believe the primary way of utilizing IDOM should be embedding within existing applications.

- #654

### Proposed Actions

- [ ] Keep `idom.run()` as is, but add logging to indicate `idom.run()` is not suited for production use.
   - ex. `idom.run() is only intended for testing purposes. For production use, consider embedding IDOM within an existing project. Check out the docs for more info.`
- [ ] Expand the ease-of-use for integrating into supported web frameworks, similar to Django-IDOM
   - #653
- [ ] Remove SharedClientState, as it is not compatible with multiprocessed webservers.
   - This is also an opportunity to rename `PerClientState` to `IdomView` 

#### Optional
- Change the callable name/path to more appropriately indicate **danger** or **for testing purposes only**
   - `idom.testing.webserver()`
   - `idom.shortcuts.run()`
   - `idom.devtools.run_webserver()`
- Remove the pip parameter for frameworks (it's really not needed)
   - `idom.run()` should raise a simple to understand exception if the user has forgotten to install a web framework
   - ex. `Looks like you don't have a web framework installed. You'll need to pip install one of the following: fastapi, flask, sanic`

Archmonger:
My personal opinion is that `idom.run()` should be tucked away into a "dev tools" closet as-is and shouldn't be prominently advertised.

Instead, we should focus on simplifying how users integrate IDOM into existing applications.

rmorshea:
We could add a big fat warning about this instead? Technically doing `idom.run` is fine with Sanic (and I think Tornado too), but not  the others.

rmorshea:
Meh, nevermind. We should completely shift how this works:

- Drop the run function
- stop calling things a per-client state "server" and instead call it a "view"
- Make `app` a required argument of `PerClientStateView` etc.

Doing this with appropriate deprecation warnings will be a bit tricky, but I think that sends a more accurate message about how you should expect to use IDOM.

rmorshea:
The only question then is how you many the "hello world" still look simple...

rmorshea:
Maybe do:

```python
from idom import component, html, server


@component
def App():
    return html.h1("Hello, World!")


[...]

Issue #669, opened 2022-02-15
Provide access to RequestContext 


Archmonger:
### Current Situation

Currently, only Django-IDOM has access to the current websocket's `scope`. However, this is very commonly needed within all web frameworks to provide access to things such as user attributes, login, and logout functionality.

See https://github.com/idom-team/idom/discussions/594

### Proposed Actions

Create a `RequestContext` to be used with the upcoming `use_context` hook.

### Work Items

- Create `RequestContext` on a per web-framework basis
- Document this feature (Probably also want an example on user login/logout)

Issue #591, opened 2022-01-15
Use Async WebDiver in Tests


rmorshea:
### Current Situation

Currently, we run tests which rely on the browser using a synchronously executing web driver from Selenium. As a result, server implementations have been required to provide a `run_in_thread` method such that they can be run in the background while the blocking test code is allowed to proceed.

Unfortunately, some async servers don't play well with threads. This is primarily due to the fact that cleanly stopping a thread can be somewhat challenging. For example, IDOM is currently locked into version `sanic<19.12` because later releases have been developed without consideration for threading.

### Proposed Actions

As a result, we should switch to an async web driver like [arsenic](https://github.com/HDE/arsenic) or [pyppeteer](https://miyakogi.github.io/pyppeteer/reference.html) and add a `ServerType.run_async` method to the various server implementations. Once this is done we can deprecate the `ServerType.run_in_thread` method.

Then lastly, we can upgrade Sanic to the latest version.

### Work Items

- [ ] Implement `ServerType.run_async` and switch to async web driver
- [ ] Mark `ServerType.run_in_thread` as deprecated
- [ ] Update downstream libraries (e.g. custom component template and anything made from it).
- [ ] Upgrade `sanic` and raise when `run_in_thread` is called for `sanic>=19.12`
- [ ] Remove `ServerType.run_in_thread`

Archmonger:
We can also use [arsenic](https://github.com/HDE/arsenic) instead of Selenium.

It would save us a lot of effort in developing custom tooling.

rmorshea:
This is somewhat more urgent as a result of: https://github.com/idom-team/idom/security/dependabot/9

Archmonger:
Security issues are not publicly visible, and I don't have access to them.

rmorshea:
I edited the link

PR #703, opened 2022-03-06
Rework How IDOM Integrates With Servers

rmorshea:
closes: #657

The change proposed here is that IDOM will expose a top-level `idom.develop` function that takes in a component definition and not much else. The configuration options here are intentionally limited since we want to push people away from using this function in production.

This change will also remove `PerClientStateServer` and `SharedClientStateServer` with a single `configure` function for each server implementation:

```python
from idom.server.sanic import configure
configure(app, MyComponent)
```

As part of this change `SharedClientStateServer` will be removed. While it's cool, it can't be used in a production context - you need a real cross-process solution for synchronizing state (e.g. Redis).

## Other issues I plan to address here:

closes: #678 #591 #669 #657


Archmonger:
Will need to keep in mind that readme and docs modifications will need to occur here too. 

I don't know if I agree with adding an app parameter to `idom.run` since it uglifies how new user test IDOM. I think a warning message will suffice.

Archmonger:
Here's an alternative syntax to consider, based more identically on [WhiteNoise's design](http://whitenoise.evans.io/en/stable/base.html#using-whitenoise-with-any-wsgi-application)

```python
from sanic import Sanic
from idom.server.sanic import IdomSanic

# Configure returns a wrapped ASGI/WSGI callable
asgi_app = IdomSanic(Sanic("My App"), MyComponent, endpoint="/idom")
```

rmorshea:
I don't think IDOM has a particularly compelling reason to create provide an ASGI app, so for now I think sticking with a simple `configure` function should be fine.

Archmonger:
The only compelling reason is if you want IDOM to behave more like WSGI/ASGI middleware.

Side note, I'm assuming the reason the configure function exists is to set up the WS/HTTP endpoints, as mentioned in #653 ?

rmorshea:
By "compelling reason", I mean that I can't think of any methods or middleware to add to an IDOM ASGI app.

And yes, the recommended way of running IDOM would be to instantiate a support application, `configure()` it, and then run the app with an ASGI server.

Archmonger:
Not pushing for this change, both design patterns accomplish the same task. But on the topic of middleware, it doesn't necessarily need to add methods. See [`Whitenoise().__call__`](https://github.com/evansd/whitenoise/blob/ac902ecb9fa0625d4e3cf66c6f9a9976dd7cace2/src/whitenoise/base.py#L78).

rmorshea:
I'm just not sure what IDOM would want to do in the `__call__` method of its ASGI app.

I think we can stick with the `configure` method for now. If it turns out we want to create an ASGI wrapper it should be pretty straightforward to do so. We might not even need to deprecate the `configure` function since the ASGI app might continue to call that.

PR #710, opened 2022-03-17
Update running-idom.rst

jmtaysom:
The run function does not have a `config` kwarg but instead `server_config`

### DIFF
--- a/docs/source/guides/getting-started/running-idom.rst
+++ b/docs/source/guides/getting-started/running-idom.rst
@@ -157,7 +157,7 @@ then be passed to the ``run()`` function via the ``config`` parameter:
         redirect_root_to_index=True,
     )
 
-    run(App, server_type=PerClientStateServer, config=server_config)
+    run(App, server_type=PerClientStateServer, server_config=server_config)
 
 Here's the list of available configuration types:
 


rmorshea (approved):

rmorshea:
Thanks @jmtaysom, a note on this, we're thinking about making [some changes](https://github.com/idom-team/idom/issues/657) to the `run` function since it encourages some bad habits - namely that you should use a production [WSGI](https://wsgi.readthedocs.io/en/latest/servers.html) or [ASGI](https://asgi.readthedocs.io/en/latest/implementations.html) web server instead of the `run` function. In the future `idom.run` will become a tool for convenience while developing rather than something that should be used in production.

jmtaysom:
I saw some of the chatter in the issues about that. I am definitely all for
that. I have also had issues with the shared runtime in the tutorial. So
that may be a section to clean up in the future if it doesn't get cleaned
up automatically by moving away from the run command.

rmorshea:
> I have also had issues with the shared runtime in the tutorial

Can you link to particular sections of the docs that you found to be problematic? I'll create an issue to track that.

jmtaysom:
https://idom-docs.herokuapp.com/docs/guides/getting-started/running-idom.html?highlight=server%20name#available-server-types

When I ran the above section about SharedClientStateServer I get the
following error
sanic.exceptions.SanicException: Sanic instance named
'idom.server.sanic[1]' uses an invalid format. Names must begin with a
character and may only contain alphanumeric characters, _, or -.

.I ran the same section with an explicitly created and named sanic app and
passed that app into the run function which let it work but I still
couldnt get two windows to update each other as I moved the slider

rmorshea:
It looks like there may be two separate problems here. I'll open a discussion so we can tease them out.

PR #739, opened 2022-04-16
another readme rewrite

rmorshea:
Yet another take on the README. Not sure I like it yet, but I think it's more inline with how I imagine it looking in the end.

### DIFF
--- a/src/idom/_option.py
+++ b/src/idom/_option.py
@@ -107,6 +107,7 @@ class DeprecatedOption(Option[_O]):  # pragma: no cover
         warnings.warn(
             f"{self.name!r} has been renamed to {self._new_opt.name!r}",
             DeprecationWarning,
+            stacklevel=3,
         )
         return self._new_opt.current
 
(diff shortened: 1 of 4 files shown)


src/idom/_option.py
```diff
@@ -107,6 +107,7 @@ def _current(self) -> _O:  # type: ignore
         warnings.warn(
             f"{self.name!r} has been renamed to {self._new_opt.name!r}",
             DeprecationWarning,
+            stacklevel=3,
```
rmorshea, line 110: Not related, just something I and wanted to improve caught.

Archmonger (changes requested):
I personally dislike this one. It really doesn't give a top-level highlight of the main reasons why anyone would ever want to use IDOM.

The main thing about the readme is trying to "sell" IDOM to people who haven't read the docs yet. We need to highlight all our selling points as concisely as possible. 90% of people will never read the docs.

In my previous PR, I listed the main things that I consider selling points for new developers.

Also, we definitely want to avoid tech jargon in any readme we create. Words such as "declarative" shouldn't make it in.

rmorshea:
Good point.

rmorshea:
I think the README more directly gets at the features IDOM provides, but it's also more scattershot now. Maybe there's too much that we're trying to fit into it?

Rather than throwing all the features at the reader at once we focus on something like 3:

- Write apps without JS
- Integration with web frameworks like FastAPI, Django, etc
- React-style components in Python

@acivitillo or @liberty-rising do you have some thoughts on what the reader our should come away with from the README?

rmorshea:
I'm gonna close this and open a discussion.

PR #746, opened 2022-04-23
added selectedOptions to select onchange event

wrabit:
I noticed the event payload from a `<select>` element that has `multiple="true"` doesn't contain `selectedOptions` property so it's impossible to know all of the items that are selected.


rmorshea:
Thanks for the contribution! If you could add a test for this [here](https://github.com/idom-team/idom/blob/main/src/client/packages/idom-client-react/tests/event-to-object.test.js) that would be great.

wrabit:
selectedOptions was not in-fact part of the change event but rather a property of the select element. Will have to tackle a different way. Is there a simpler way (than creating a javascript / react module) to get this? All we would need is something like:

```
let values = []; 

for (let item of [element].selectedOptions) {
    values.push(item.value)
}
```

PR #770, opened 2022-06-22
use standard context provider pattern

rmorshea:
# Description

The current implementation has `create_context` return a new `Context` class. This makes extending `Context` more complicated since you need to create a new subclass in addition to the one you might return from a `create_my_context` function in a similar way to how `create_context` works now. For example:

```python
from idom.core.hooks import Context

class MyContext(Context):
    custom_attribute = "something"

def create_my_context(custom_attribute):
    class _MyContext(Context):
        custom_attribute = custom_attribute
    _MyContext.__name__ = "MyContext"
    return MyContext
```

This is just overly complicated. This rework makes the following, much simpler extension, possible:

```python
from idom.core.hooks import ContextType

class MyContextType:
    def __init__(self, custom_attribute):
        super().__init__()

def create_my_context(custom_attribute):
    return MyContextType(custom_attribute)
```

# Checklist:

- [x] I have left irrelevant checklist items unchecked instead of removing them.

### DIFF
--- a/src/idom/backend/flask.py
+++ b/src/idom/backend/flask.py
@@ -26,7 +26,7 @@ from werkzeug.serving import BaseWSGIServer, make_server
 
 import idom
 from idom.backend.types import Location
-from idom.core.hooks import Context, create_context, use_context
+from idom.core.hooks import ContextType, create_context, use_context
 from idom.core.layout import LayoutEvent, LayoutUpdate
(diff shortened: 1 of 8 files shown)


rmorshea:
On second thought, it's more complicated, but people probably shouldn't be extending context anyway. Best for people to just wrap it instead.

PR #787, opened 2022-07-11
Track contexts in hooks as state

rmorshea:
# Description

closes: #789 

components which are newly added to the layout after the initial render do not have access to contexts. we solve this by tracking life cycle hooks in stack and copying the context providers from parent to child hooks.

# Checklist:

### DIFF
--- a/docs/source/about/changelog.rst
+++ b/docs/source/about/changelog.rst
@@ -23,13 +23,18 @@ more info, see the :ref:`Contributor Guide <Creating a Changelog Entry>`.
 Unreleased
 ----------
 
-**Added**
+**Fixed**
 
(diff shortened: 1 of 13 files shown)


rmorshea:
@Archmonger, thoughts on removing `create_context` in favor of a simple `Context` object constructor?

Archmonger:
I am completely indifferent to that change. At the moment, you and I are the only ones using it. And I don't expect many people to create custom context hooks for IDOM in the future.

rmorshea:
I decided not to do that. Came up with a nice solution using functions. I'm going to remove the `name` parameter of `create_context` though. React doesn't have it, and I'm not sure it provides that much value.

PR #795, opened 2022-08-04
Use LXML for html_to_vdom

### LINKED ISSUE #777: html_to_vdom breaks on whitespaces and null HTML tags
### Current Situation

If a HTML string has ...

- Additional whitespace at the tail of the HTML document
- A null (non-terminated) HTML tag such as `<hr>` within the document

... then `html_to_vdom` will silently fail without raising an exception

See [my Django IDOM PR](https://github.com/idom-team/django-idom/pull/81) for how I caused the `script` tag issue to occur at [this LOC](https://github.com/Archmonger/django-idom/blob/1dda52d01022cca47da25610c2cf97a45c6a34d3/tests/test_app/components.py#L144).

### Proposed Actions

Whitespace issue can be fixed by simple regex and/or using `str.strip()`.

Script tag issue needs further investigation. This might open up a can of worms to test `html_to_vdom` with all `idom.html` types.

Archmonger:
fix #777

Technical Details
- Generate a DOM tree using lxml
- Recursively converts lxml tree to VDOM
- HTML is encapsulated within a React fragment if needed
- Style tags are handled separately
- Utilize `idom.html` constructors for VDOM when possible
- Unneeded VDOM fields are pruned after user transformations are applied
- Uses list comprehension where possible for maximum performance

## Checklist

### DIFF
--- a/src/idom/utils.py
+++ b/src/idom/utils.py
@@ -112,7 +112,7 @@ class _HtmlToVdom:
             vdom["tagName"] = ""
 
         # Get rid of empty VDOM fields
-        self._remove_unneeded_vdom(vdom)
+        self._prune_vdom_fields(vdom)
 
         return vdom
 
@@ -140,7 +140,7 @@ class _HtmlToVdom:
                 vdom["key"] = vdom["children"][0]
 
     @staticmethod
-    def _remove_unneeded_vdom(vdom: Dict):
+    def _prune_vdom_fields(vdom: Dict):
         """Removed unneeded fields from VDOM dict."""
         if not len(vdom["children"]):
             del vdom["children"]
--- a/src/idom/utils.py
+++ b/src/idom/utils.py
@@ -119,18 +119,17 @@ def _etree_to_vdom(node: etree._Element, transforms: Iterable[_ModelTransform]) 
     # Convert the lxml node to a VDOM dict
     attributes = dict(node.items())
     key = attributes.pop("key", None)
-    vdom = (
-        # If available, use a constructor from idom.html to create the VDOM dict
-        getattr(idom.html, node.tag)(attributes, *children, key=key)
-        if hasattr(idom.html, node.tag)
-        # Fall back to using a generic VDOM dict
-        else {
-            "tagName": node.tag,
-            "children": children,
-            "attributes": attributes,
-            "key": key,
-        }
-    )
+
+    if hasattr(idom.html, node.tag):
+        vdom = getattr(idom.html, node.tag)(attributes, *children, key=key)
+    else:
+        vdom: VdomDict = {"tagName": node.tag}
+        if children:
+            vdom["children"] = children
+        if attributes:
+            vdom["attributes"] = attributes
+        if key is not None:
+            vdom["key"] = key
 
     # Perform any necessary mutations on the VDOM attributes to meet VDOM spec
     _mutate_vdom(vdom)
--- a/src/idom/utils.py
+++ b/src/idom/utils.py
@@ -67,9 +67,10 @@ def html_to_vdom(html: str, *transforms: _ModelTransform):
         raise TypeError(f"Encountered unsupported type {type(html)} from {html}")
 
     # If the user provided a string, convert it to a list of lxml.etree nodes
-    nodes: List = fragments_fromstring(
-        html, no_leading_text=True, parser=etree.HTMLParser()
+    parser = etree.HTMLParser(
+        remove_comments=True, remove_pis=True, remove_blank_text=True
     )
+    nodes: List = fragments_fromstring(html, no_leading_text=True, parser=parser)
     has_root_node = len(nodes) == 1
 
     # Find or create a root node
--- a/src/idom/utils.py
+++ b/src/idom/utils.py
@@ -53,7 +53,7 @@ class Ref(Generic[_RefValue]):
         return f"{type(self).__name__}({current})"
 
 
-def html_to_vdom(html: str, *transforms: _ModelTransform):
+def html_to_vdom(html: str, *transforms: _ModelTransform, recover: bool = False) -> Dict:
     """Transform HTML into a DOM model
     Parameters:
         source:
@@ -62,13 +62,19 @@ def html_to_vdom(html: str, *transforms: _ModelTransform):
             Functions of the form ``transform(old) -> new`` where ``old`` is a VDOM
             dictionary which will be replaced by ``new``. For example, you could use a
             transform function to add highlighting to a ``<code/>`` block.
+        recover:
+            If ``True``, try to repair broken HTML. This may result in parsing invalid
+            HTML as plain text.
     """
     if not isinstance(html, str):
         raise TypeError(f"Encountered unsupported type {type(html)} from {html}")
 
     # If the user provided a string, convert it to a list of lxml.etree nodes
     parser = etree.HTMLParser(
-        remove_comments=True, remove_pis=True, remove_blank_text=True
+        remove_comments=True,
+        remove_pis=True,
+        remove_blank_text=True,
+        recover=recover,
     )
     nodes: List = fragments_fromstring(html, no_leading_text=True, parser=parser)
     has_root_node = len(nodes) == 1
--- a/src/idom/utils.py
+++ b/src/idom/utils.py
@@ -5,6 +5,7 @@ from lxml import etree
 from lxml.html import fragments_fromstring
 
 import idom
+from idom.core.vdom import VdomDict
 
 
 _RefValue = TypeVar("_RefValue")
@@ -55,7 +56,7 @@ class Ref(Generic[_RefValue]):
         return f"{type(self).__name__}({current})"
 
 
-def html_to_vdom(html: str, *transforms: _ModelTransform) -> Dict:
+def html_to_vdom(html: str, *transforms: _ModelTransform) -> VdomDict:
     """Transform HTML into a DOM model. Unique keys can be provided to HTML elements
     using a ``key=...`` attribute within your HTML tag.
 
@@ -99,7 +100,9 @@ def html_to_vdom(html: str, *transforms: _ModelTransform) -> Dict:
     return vdom
 
 
-def _etree_to_vdom(node: etree._Element, transforms: Iterable[_ModelTransform]) -> Dict:
+def _etree_to_vdom(
+    node: etree._Element, transforms: Iterable[_ModelTransform]
+) -> VdomDict:
     """Recusively transform an lxml etree node into a DOM model
 
     Parameters:
@@ -141,7 +144,7 @@ def _etree_to_vdom(node: etree._Element, transforms: Iterable[_ModelTransform]) 
     return vdom
 
 
-def _mutate_vdom(vdom: Dict):
+def _mutate_vdom(vdom: VdomDict):
     """Performs any necessary mutations on the VDOM attributes to meet VDOM spec.
 
     Currently, this function only transforms the ``style`` attribute into a dictionary whose keys are
@@ -168,7 +171,7 @@ def _mutate_vdom(vdom: Dict):
 
 def _generate_vdom_children(
     node: etree._Element, transforms: Iterable[_ModelTransform]
-) -> List[Union[Dict, str]]:
+) -> List[Union[VdomDict, str]]:
     """Generates a list of VDOM children from an lxml node.
 
     Inserts inner text and/or tail text inbetween VDOM children, if necessary.
--- a/src/idom/utils.py
+++ b/src/idom/utils.py
@@ -53,7 +53,7 @@ class Ref(Generic[_RefValue]):
         return f"{type(self).__name__}({current})"
 
 
-def html_to_vdom(html: str, *transforms: _ModelTransform, recover: bool = False) -> Dict:
+def html_to_vdom(html: str, *transforms: _ModelTransform) -> Dict:
     """Transform HTML into a DOM model
     Parameters:
         source:
@@ -62,9 +62,6 @@ def html_to_vdom(html: str, *transforms: _ModelTransform, recover: bool = False)
             Functions of the form ``transform(old) -> new`` where ``old`` is a VDOM
             dictionary which will be replaced by ``new``. For example, you could use a
             transform function to add highlighting to a ``<code/>`` block.
-        recover:
-            If ``True``, try to repair broken HTML. This may result in parsing invalid
-            HTML as plain text.
     """
     if not isinstance(html, str):
         raise TypeError(f"Encountered unsupported type {type(html)} from {html}")
@@ -74,7 +71,7 @@ def html_to_vdom(html: str, *transforms: _ModelTransform, recover: bool = False)
         remove_comments=True,
         remove_pis=True,
         remove_blank_text=True,
-        recover=recover,
+        recover=False,
     )
     nodes: List = fragments_fromstring(html, no_leading_text=True, parser=parser)
     has_root_node = len(nodes) == 1
@@ -97,7 +94,7 @@ def html_to_vdom(html: str, *transforms: _ModelTransform, recover: bool = False)
     return vdom
 
 
-def _etree_to_vdom(node: etree._Element, *transforms: _ModelTransform):
+def _etree_to_vdom(node: etree._Element, *transforms: _ModelTransform) -> Dict:
     """Recusively transform an lxml etree node into a DOM model
     Parameters:
         source:
--- a/src/idom/utils.py
+++ b/src/idom/utils.py
@@ -64,16 +64,14 @@ def html_to_vdom(html: Union[str, etree._Element], *transforms: _ModelTransform)
             transform function to add highlighting to a ``<code/>`` block.
     """
 
-    # Keep track of whether this is the root node
-    root_node = False
-
     # If the user provided a string, convert it to an lxml.etree node.
     if isinstance(html, str):
         parser = etree.HTMLParser()
         node = fragment_fromstring(html, create_parent=True, parser=parser)
-        root_node = True
+        root_node = True  # Only the root node is a HTML string
     elif isinstance(html, etree._Element):
         node = html
+        root_node = False
     else:
         raise TypeError(
             f"HtmlToVdom encountered unsupported type {type(html)} from {html}"
--- a/src/idom/utils.py
+++ b/src/idom/utils.py
@@ -82,14 +82,14 @@ def html_to_vdom(html: Union[str, etree._Element], *transforms: _ModelTransform)
     # Convert the lxml node to a VDOM dict.
     vdom = {
         "tagName": node.tag,
-        "children": _generate_child_vdom(node, transforms),
+        "children": _generate_vdom_children(node, transforms),
         "attributes": dict(node.items()),
         "eventHandlers": {},
         "importSource": {},
         "key": "",
         "error": "",
     }
-    _vdom_mutations(vdom)
+    _mutate_vdom(vdom)
 
     # Apply any provided transforms.
     for transform in transforms:
@@ -105,7 +105,7 @@ def html_to_vdom(html: Union[str, etree._Element], *transforms: _ModelTransform)
     return vdom
 
 
-def _vdom_mutations(vdom: Dict):
+def _mutate_vdom(vdom: Dict):
     """Performs any necessary mutations on the VDOM attributes to meet VDOM spec
     and/or to make elements properly renderable in React."""
     # Convert style attributes to VDOM spec
@@ -144,9 +144,9 @@ def _prune_vdom_fields(vdom: Dict):
         del vdom["error"]
 
 
-def _generate_child_vdom(
+def _generate_vdom_children(
     node: etree._Element, transforms: Iterable[_ModelTransform]
-) -> List:
+) -> List[Union[Dict, str]]:
     """Recursively generate a list of VDOM children from an lxml node."""
     # Insert text inbetween VDOM children, if necessary
     children = [node.text] + list(
--- a/src/idom/utils.py
+++ b/src/idom/utils.py
@@ -67,7 +67,7 @@ def html_to_vdom(html: str, *transforms: _ModelTransform) -> Dict:
             dictionary which will be replaced by ``new``. For example, you could use a
             transform function to add highlighting to a ``<code/>`` block.
     """
-    if not isinstance(html, str):
+    if not isinstance(html, str):  # pragma: no cover
         raise TypeError(f"Encountered unsupported type {type(html)} from {html}")
 
     # If the user provided a string, convert it to a list of lxml.etree nodes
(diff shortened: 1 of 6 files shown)


Archmonger:
@rmorshea 

This seems to work great based on my informal testing, but I'm still searching for any edge cases that may exist.

Are there any other oddities (such as how Scripts should automatically get keys) that I should look out for?

Archmonger:
Seems like our `Flask` support is broken on Python 3.10 according to the tests.

src/idom/utils.py
```diff
@@ -63,80 +63,109 @@ def html_to_vdom(source: str, *transforms: _ModelTransform) -> Dict[str, Any]:
...
-
-    def handle_data(self, data: str) -> None:
-        self._node_stack[-1]["children"].append(data)
+    if not isinstance(html, str):
+        raise TypeError("html_to_vdom expects a string!")
+
+    return _HtmlToVdom(*transforms).convert(html)
+
+
+class _HtmlToVdom:
```
rmorshea, line 73: It seems like this could be implemented with pure functions instead of a class. If so, let's do that. The only reason for the class before was that we had to subclass `HTMLParser`.

Archmonger: I did have it implemented with pure functions before, but it I converted to a class based approach since it seemed easier to read/organize.

I'm not opposed to going back to functions though, but the `utils.py` file would then be littered with one-use helper functions that exist just for internal `html_to_vdom` use.

Archmonger: I converted it back to a functional approach. Let me know which you prefer.

Archmonger:
Questions that need to be answered:

1) Should we add a `strict_parser` boolean for parsing HTML?
2) Should we automatically set a key on the root node to help transformations identify it easier?
3) Are there any other built-in VDOM mutations that need to take place, similar to `style` attribute formatting or `script` tag keys.
4) Are we certain that the old logic for converting CSS property names works properly for all use cases?
5) Is it expected behavior that `script` don't truly exist on the page, and are rather replaced with an empty `div`?
6) Is safe to use the index as the primary key for html_to_vdom components, or should we automatically default to a `uuid`?
7) Should I create a discussion thread to talk about `morphdom` vs `react-dom`?

src/idom/utils.py
```diff
@@ -49,12 +53,8 @@ def __repr__(self) -> str:
         return f"{type(self).__name__}({current})"
 
 
-_ModelTransform = Callable[[Dict[str, Any]], Any]
-
-
-def html_to_vdom(source: str, *transforms: _ModelTransform) -> Dict[str, Any]:
+def html_to_vdom(html: Union[str, etree._Element], *transforms: _ModelTransform):
```
rmorshea, line 56: To avoid exposing the underlying implementation (i.e. the `etree._Element` type), maybe there could be the public `html_to_vdom` which accepts just a string and a private `_html_to_vdom` function that does the actual recursive part of the logic.

src/idom/utils.py
```diff
@@ -63,80 +63,108 @@ def html_to_vdom(source: str, *transforms: _ModelTransform) -> Dict[str, Any]:
...
-                style_dict = {}
-                for k, v in (part.split(":", 1) for part in style.split(";") if part):
-                    title_case_key = k.title().replace("-", "")
-                    camel_case_key = title_case_key[:1].lower() + title_case_key[1:]
-                    style_dict[camel_case_key] = v
-                attrs["style"] = style_dict
-        return {"tagName": tag, "attributes": attrs, "children": []}
+    return _html_to_vdom(html, *transforms)
+
+def _html_to_vdom(html: Union[str, etree._Element], *transforms: _ModelTransform):
```
rmorshea, line 69: Should this function only accept `etree._Element`? It seems like you could put the conversion from a string to the `etree` in the body of the public `html_to_vdom` function. You could then get rid of the `if isinstance(html, str)` check here.

Archmonger: Yep definitely possible. Will do this tonight.

rmorshea:
1. I'm not really sure what "strict" implies.
2. If there's a root node in the document then just use that. If there's more than one root node I think we can just make a `tagName=""` node since they're not displayed anyway.
3. If there were a sane way to camelCase other HTML attributes we could do that. I can't think of a way though since they aren't nicely dash-separated like CSS. Regardless, I think the `style` mutation should be implemented as a transform function that we apply by default if no others are given. While the React client expects camelCase, that might not be true of all clients.
4. The script sanitization was from before we had a script element.
5. Key's should not be the responsibility of this method. Leaving the key absent (implies key is index) is safe unless the HTML was generated from a template and might change, in which case the template, or the thing rendering it is responsible for managing keys.
6. Sure. I don't recall talking about `morphdom`. Could you include some background on that?
7. Let's resolve that in a separate PR

Archmonger:
1. Strict implies that malformatted HTML will generate an exception. The default behavior of lxml is to loosely parse broken HTML.
2. It's a bit ugly to have an either/or situation (given the function is recursive), so I'd prefer if the root node is always a fragment.
3. I disagree that it should be done via the `transforms` iterable. This is likely to result in a lot of user headache if they attempt to use our `transforms` without realizing they'd immediately lose the `style` transforms.
4. Is this a bug that needs to be fixed then?
5. N/A
6. See [the comment above](https://github.com/idom-team/idom/pull/795#issuecomment-1204951752)
7. N/A

Archmonger:
`etree_to_vdom` was added to be the "recursive stuff". It _technically_ can be documented, but I don't know if we want to. If we do, we're indirectly telling people it's normal to create DOM trees in `lxml`, which doesn't feel right.

I will keep it as an undocumented utility unless you feel otherwise.

rmorshea:
It seems like it should be private then. Can you rename it to `_etree_to_vdom`?

rmorshea:
1. I think we can probably leave out the `recover` flag you've added and parse the HTML strictly. If someone needs it to be loosely parsed we can add in the flag later.
3. Seems reasonable.
4. Not a bug, just an old design decision. Now that IDOM's client supports script elements we can stop stripping them.
6. I think we'd write an entirely new client if we were to use `morphdom`. Very little of the React-based client code could be re-used. In order to justify spending time on that we'd need to do a bunch of benchmarking.

src/idom/utils.py
```diff
@@ -63,80 +63,130 @@ def html_to_vdom(source: str, *transforms: _ModelTransform) -> Dict[str, Any]:
...
+        del vdom["key"]
+    if not vdom["error"]:
+        del vdom["error"]
+
+
+def _generate_vdom_children(
+    node: etree._Element, transforms: Iterable[_ModelTransform]
+) -> List[Union[Dict, str]]:
+    """Recursively generate a list of VDOM children from an lxml node.
+    Inserts inner text and/or tail text inbetween VDOM children, if necessary."""
```
rmorshea, line 175:
```suggestion
    """Recursively generate a list of VDOM children from an lxml node.
    
    Inserts inner text and/or tail text inbetween VDOM children, if necessary.
    """
```

src/idom/utils.py
```diff
@@ -63,80 +63,130 @@ def html_to_vdom(source: str, *transforms: _ModelTransform) -> Dict[str, Any]:
...
+    # Change the artificially created root node to a React Fragment, instead of a div
+    if not has_root_node:
+        vdom["tagName"] = ""
+
+    return vdom
+
+
+def _etree_to_vdom(node: etree._Element, *transforms: _ModelTransform) -> Dict:
+    """Recusively transform an lxml etree node into a DOM model
+    Parameters:
```
rmorshea, line 99:
```suggestion
    """Recusively transform an lxml etree node into a DOM model
    
    Parameters:
```

src/idom/utils.py
```diff
@@ -63,80 +63,130 @@ def html_to_vdom(source: str, *transforms: _ModelTransform) -> Dict[str, Any]:
...
+
+    # Get rid of empty VDOM fields
+    _prune_vdom_fields(vdom)
+
+    return vdom
+
+
+def _mutate_vdom(vdom: Dict):
+    """Performs any necessary mutations on the VDOM attributes to meet VDOM spec
+    and/or to make elements properly renderable in React."""
```
rmorshea, line 134:
```suggestion
    """Performs any necessary mutations on the VDOM attributes to meet VDOM spec.
    
    It also transforms the ``style`` attribute into a dictionary whose keys are
    camelCase so as to be renderable by React.
    """
```

src/idom/utils.py
```diff
@@ -63,80 +63,130 @@ def html_to_vdom(source: str, *transforms: _ModelTransform) -> Dict[str, Any]:
...
+    # Convert the lxml node to a VDOM dict.
+    vdom = {
+        "tagName": node.tag,
+        "children": _generate_vdom_children(node, transforms),
+        "attributes": dict(node.items()),
+        "eventHandlers": {},
+        "importSource": {},
+        "key": "",
+        "error": "",
+    }
```
rmorshea, line 119: If the `hasattr(idom.html, tagName)`, let's use the constructors from that module. This will give useful errors for elements which are not allowed to have children as well has handle any extra logic (e.g. the `script` element). For elements not in the `idom.html` module, you'll only need to declare the `tagName`,  `children`, `attributes`, and `key` fields - the others cannot be defined via static HTML at the moment.

For the `key` of an element, you'll want to pop that off the attributes dict and assign that to the VDOM dict, or hand it to an `html` constructor, separately - this was not properly handled in the old version either.

src/idom/utils.py
```diff
@@ -63,80 +63,130 @@ def html_to_vdom(source: str, *transforms: _ModelTransform) -> Dict[str, Any]:
...
+            )
+        }
+
+    # Set key attribute for scripts to prevent re-execution during re-renders
+    if vdom["tagName"] == "script":
+        if not isinstance(vdom["children"][0], str):
+            # The script's source should always be the first child
+            raise LookupError("Could not find script's contents!")
+        if vdom["children"][0]:
+            vdom["key"] = vdom["children"][0]
```
rmorshea, line 152: This can be removed if we use `idom.html` constructors as suggested above.

rmorshea:
There's a problem with the editable install on Python 3.10.6 that should be fixed in my PR.

src/idom/utils.py
```diff
@@ -63,80 +67,149 @@ def html_to_vdom(source: str, *transforms: _ModelTransform) -> Dict[str, Any]:
...
+    first_word, *subsequent_words = css_key.split("-")
+
+    return "".join(
+        [
+            # Lowercase the first word
+            first_word.lower(),
+            # Use map() to titlecase all subsequent words
+            *map(str.title, subsequent_words),
+        ]
+    )
```
rmorshea, line 215:
```suggestion
def _hypen_to_camel_case(string: str) -> str:
    """Convert a hypenated string to camelCase."""
    first, remainder = string.split("-", 1)
    return first.lower() + remainder.title().replace("-", "")
```

src/idom/utils.py
```diff
@@ -63,80 +67,149 @@ def html_to_vdom(source: str, *transforms: _ModelTransform) -> Dict[str, Any]:
...
-
-            parser = HtmlParser()
-
-            parser.feed(an_html_string)
-            parser.feed(another_html_string)
-            ...
-
-            vdom = parser.model()
+    if not isinstance(html, str):
+        raise TypeError(f"Encountered unsupported type {type(html)} from {html}")
```
rmorshea, line 71:
```suggestion
        raise TypeError(f"Expected html to be a string, not {type(html).__name__}")
```

src/idom/utils.py
```diff
@@ -63,80 +67,149 @@ def html_to_vdom(source: str, *transforms: _ModelTransform) -> Dict[str, Any]:
...
+        recover=False,
+    )
+    nodes: List = fragments_fromstring(html, no_leading_text=True, parser=parser)
+    has_root_node = len(nodes) == 1
+
+    # Find or create a root node
+    if has_root_node:
+        root_node = nodes[0]
+    else:
+        root_node = etree.Element("div", None, None)
```
rmorshea, line 87:
```suggestion
        # etree.Element requires a non-empty tag - we correct this below
        root_node = etree.Element("TEMP", None, None)
```

src/idom/utils.py
```diff
@@ -63,80 +67,149 @@ def html_to_vdom(source: str, *transforms: _ModelTransform) -> Dict[str, Any]:
...
+    vdom = _etree_to_vdom(root_node, transforms)
+
+    # Change the artificially created root node to a React Fragment, instead of a div
+    if not has_root_node:
+        vdom["tagName"] = ""
+
+    return vdom
+
+
+def _etree_to_vdom(node: etree._Element, transforms: Iterable[_ModelTransform]) -> Dict:
```
rmorshea, line 101: You can use `idom.core.vdom.VdomDict` to specify this and other types here.

rmorshea (changes requested):

src/idom/utils.py
Archmonger: Calling `__name__` on a `type` call is risky, since it does not handle `None`

Archmonger:
@rmorshea force pushing on a remote fork typically isn't a good idea, since it causes the owner of the fork to have to delete and repull his local branches.

GitHub automatically hides all commits that are within the parent branch of a PR (`idom-team:main`), so would've been better to do a simple merge here.

src/idom/utils.py
```diff
@@ -63,80 +67,141 @@ def html_to_vdom(source: str, *transforms: _ModelTransform) -> Dict[str, Any]:
...
+    Parameters:
+        source:
+            The ``lxml.etree._Element`` node
+        transforms:
+            Functions of the form ``transform(old) -> new`` where ``old`` is a VDOM
+            dictionary which will be replaced by ``new``. For example, you could use a
+            transform function to add highlighting to a ``<code/>`` block.
     """
+    if not isinstance(node, etree._Element):
+        raise TypeError(f"Encountered unsupported type {type(node)} from {node}")
```
rmorshea, line 114: I think you can safely ignore coverage here. But then again, you could probably skip this check entirely given that the function is private.

src/idom/utils.py
Archmonger: This causes an exception for CSS variables without a dash such as `color` or `background`.

src/idom/utils.py
```diff
@@ -63,80 +67,141 @@ def html_to_vdom(source: str, *transforms: _ModelTransform) -> Dict[str, Any]:
...
+        getattr(idom.html, node.tag)(attributes, *children, key=key)
+        if hasattr(idom.html, node.tag)
+        # Fall back to using a generic VDOM dict
+        else {
+            "tagName": node.tag,
+            "children": children,
+            "attributes": attributes,
+            "key": key,
+        }
+    )
```
rmorshea, line 133: The built-in constructors already prune empty fields so, if you only assign fields whose values are not empty, you don't need `_prune_vdom_fields` anymore.

```suggestion
    attributes = dict(node.items())
    key = attributes.pop("key", None)

    if hasattr(idom.html, node.tag):
        vdom = getattr(idom.html, node.tag)(attributes, *children, key=key)
    else:
        vdom = {"tagName": node.tag}
        if children:
            vdom["children"] = children
        if attributes:
            vdom["attributes"] = attributes
        if key is not None:
            vdom["key"] = key
```

src/idom/utils.py
```diff
@@ -63,80 +67,141 @@ def html_to_vdom(source: str, *transforms: _ModelTransform) -> Dict[str, Any]:
...
-            if isinstance(style, str):
-                style_dict = {}
-                for k, v in (part.split(":", 1) for part in style.split(";") if part):
-                    title_case_key = k.title().replace("-", "")
-                    camel_case_key = title_case_key[:1].lower() + title_case_key[1:]
-                    style_dict[camel_case_key] = v
-                attrs["style"] = style_dict
-        return {"tagName": tag, "attributes": attrs, "children": []}
+    # Get rid of empty VDOM fields
+    _prune_vdom_fields(vdom)
```
rmorshea, line 143: Can be removed based on comment above.
```suggestion
```

rmorshea (changes requested):

src/idom/utils.py
Archmonger: Using `VdomDict` adds type hint errors

image

rmorshea: I think this can be fixed with the comment below updated suggestion below. Usually, you have to declare a dict literal as a `VdomDict` when it's first constructed.

src/idom/utils.py
Archmonger: This proposed solution isn't using list comprehension. So it will be slower at scale.

rmorshea: I don't see a list comprehension in the original implementation, just an inline if statement. Commented in wrong location?

src/idom/utils.py
rmorshea: Good point. This will work though:
```suggestion
def _hypen_to_camel_case(string: str) -> str:
    """Convert a hypenated string to camelCase."""
    first, _, remainder = string.partition("-")
    return first.lower() + remainder.title().replace("-", "")
```

src/idom/utils.py
Archmonger: Yep

src/idom/utils.py
rmorshea: It seems fairly reliable to say it's present and a string since the constructor for custom classes raises when `type(None, (object,), {})` because the new class' name must be a string. I don't think I've ever seen a class without a `__name__` before. Do you know of any examples?

src/idom/utils.py
Archmonger: This is the remaining error
image

`VdomAttributes` seems to be type hinted as a `Mapping[str, Any]`, which apparently doesn't have `__setitem__` defined.

src/idom/utils.py
rmorshea: Also `type(None).__name__ == "NoneType"`

src/idom/utils.py
```diff
@@ -63,80 +68,130 @@ def html_to_vdom(source: str, *transforms: _ModelTransform) -> Dict[str, Any]:
...
+    ):
+        # Convert style attribute from str -> dict with camelCase keys
+        vdom["attributes"]["style"] = {
+            _hypen_to_camel_case(key.strip()): value.strip()
+            for key, value in (
+                part.split(":", 1)
+                for part in vdom["attributes"]["style"].split(";")
+                if ":" in part
+            )
+        }
```
rmorshea, line 169: Presently the types for `VdomDict` try to be immutable where possible. Whether we should allow them to be mutable can revisited later.

```suggestion
attributes = vdom.get("attributes")
    if (
        attributes
        and "style" in attributes
        and isinstance(attributes["style"], str)
    ):
        # Convert style attribute from str -> dict with camelCase keys
        vdom["attributes"] = {
            **attributes,
            "style": {
                hypen_to_camel_case(key.strip()): value.strip()
                for key, value in (
                    part.split(":", 1)
                    for part in vdom["attributes"]["style"].split(";")
                    if ":" in part
                )
            }
        }
```

Archmonger: Recursively recreating every styled node like this really sucks, especially for DOM trees with a lot of nodes that are all styled.

Additionally, I don't think there's inherently a reason for VdomDict to be immutable?

src/idom/utils.py
Archmonger: True about the `type(None).__name__`. I was thinking of `None.__name__`, which is an exception.

src/idom/utils.py
rmorshea: To clarify, `VdomAttributes` can be mutable. It's just that it doesn't have to be. Thus, MyPy is complaining here because, in the case it were not mutable, it would not be safe to assume we can assign to it. So it would be more accurate to say that `VdomAttribute` may be immutable.

Instead of the solution above, it would be better to just convince MyPy that what we're doing is in fact safe using an `assert` statement:

```suggestion
    if (
        "attributes" in vdom
        and "style" in vdom["attributes"]
        and isinstance(vdom["attributes"]["style"], str)
    ):
        # convince type checker that it's safe to mutate attributes
        assert isinstance(vdom["attributes"], dict)
        # Convert style attribute from str -> dict with camelCase keys
        vdom["attributes"]["style"] = {
            _hypen_to_camel_case(key.strip()): value.strip()
            for key, value in (
                part.split(":", 1)
                for part in vdom["attributes"]["style"].split(";")
                if ":" in part
            )
        }
```

tests/test_utils.py
```diff
@@ -89,10 +86,49 @@ def make_links_blue(node):
                 "children": ["universe"],
                 "attributes": {"style": {"color": "blue"}},
             },
+            "lmao",
+        ],
+    }
+
+    assert html_to_vdom(source, make_links_blue) == expected
+
```
rmorshea, line 94: This should get the last bit of coverage.

```suggestion


def test_non_html_tag_behavior():
    source = "<my-tag data-x=something><my-other-tag key=a-key /></my-tag>"

    expected = {
        "tagName": "my-tag",
        "attributes": {"data-x": "something"},
        "children": [
            {"tagName": "my-other-tag", "key": "a-key"},
        ],
    }

    assert html_to_vdom(source) == expected

```

tests/test_utils.py
```diff
@@ -89,10 +86,63 @@ def make_links_blue(node):
...
+
+def test_non_html_tag_behavior():
+    source = "<my-tag data-x=something><my-other-tag key=a-key /></my-tag>"
```
rmorshea, line 97: It seems like `lxml` doesn't allow custom tags like this even though, if you define something like a web component, tags like this are valid. Maybe we should use `bs4`?

Archmonger: It would allow custom tags like this if we disable strict parsing.

rmorshea: It seems like a lot to give up the benefit of the strict parsing just to get custom tags. Does `bs4` have a strict mode that behaves similarly? If not, then maybe we expose that strict flag after all.

rmorshea:
If you want to add the strict flag back in, feel free. If not, I think we could merge as-is.

tests/test_utils.py
Archmonger: I don't recall ever seeing a toggleable option for strictness of the BS4 parser, I always assumed it was permanently lax.

I'll see if I can find any info in the docs.

src/idom/utils.py
```diff
@@ -49,11 +56,9 @@ def __repr__(self) -> str:
         return f"{type(self).__name__}({current})"
 
 
-_ModelTransform = Callable[[Dict[str, Any]], Any]
-
-
-def html_to_vdom(source: str, *transforms: _ModelTransform) -> Dict[str, Any]:
-    """Transform HTML into a DOM model
+def html_to_vdom(html: str, *transforms: _ModelTransform, strict=True) -> VdomDict:
```
rmorshea, line 59:
```suggestion
def html_to_vdom(html: str, *transforms: _ModelTransform, strict: bool = True) -> VdomDict:
```

src/idom/utils.py
```diff
@@ -62,81 +67,154 @@ def html_to_vdom(source: str, *transforms: _ModelTransform) -> Dict[str, Any]:
...
+        remove_comments=True,
+        remove_pis=True,
+        remove_blank_text=True,
+        recover=not strict,
+    )
+    try:
+        nodes: List = fragments_fromstring(html, no_leading_text=True, parser=parser)
+    except etree.XMLSyntaxError as e:
+        if not strict:
+            raise e
```
rmorshea, line 88:
```suggestion
            raise e  # pragma: no cover
```

rmorshea (approved):

tests/test_utils.py
rmorshea: we just added the strict flag + descriptive error message

rmorshea:
You should have write access now

rmorshea:
Will cut a release once this gets in.

Issue #801, opened 2022-08-12
Accidental mutation of old model causes invalid JSON Patch


rmorshea:
### Current Situation

Unfortunately, the monkey patch applied to `jsonpatch` doesn't seem to have really fixed the problem. I don't believe that we'll be able to resolve [this issue](https://github.com/stefankoegl/python-json-patch/issues/138#issuecomment-1212994774) and, even if we were, the library doesn't really seem to be maintained.

### Proposed Actions

Instead of actually computing a diff, just use a `replace` operation where the value is the new render. This will probably have a performance cost in cases where network latency dominates. The alternative though, is getting an error on occasion when JSON patch fails to generate a correct diff.

rmorshea:
This issue was discovered to still exist while working with [this test component](https://github.com/idom-team/django-idom/blob/093fc368fe974d0ee17b9c8a55169c2b6042b20e/tests/test_app/components.py#L157-L192).

rmorshea:
Turns out, the invalid JSON patch was produced because I was [mutating something](https://github.com/idom-team/idom/blob/f593ee3067cac004ca07e4c8378f87f8c45341f3/src/idom/core/layout.py#L234-L236) I shouldn't have been.

Archmonger:
Will we be sticking to using `jsonpatch` then, or are there additional critical bugs in the library?

rmorshea:
The issue I posted there is still valid, but the fix I posted for it appears to work so I don't think we need to drop `jsonpatch` in a hurry.

rmorshea:
I still think we should plan to remove it eventually though.

PR #802, opened 2022-08-12
Fix: Accidental mutation of old model causes invalid JSON Patch

rmorshea:
Closes: #801 

## Checklist


src/idom/core/layout.py
```diff
@@ -230,10 +230,17 @@ def _render_component(
...
+            old_parent_model = parent.model.current
+            old_parent_children = old_parent_model["children"]
+            parent.model.current = {
+                **old_parent_model,
+                "children": [
+                    *old_parent_children[:index],
+                    new_state.model.current,
+                    *old_parent_children[index + 1 :],
+                ],
+            }
```
Archmonger, line 243: Could this be the reason why VDOM children won't render if they follow a component?

rmorshea: It might be related to the fact that I'm replacing the parent model on line 236? The fact that I'm not propagating this change up the model parents is probably _a problem_, but I can't quite see why that's causing your problem where the children are not rendering.

Issue #806, opened 2022-09-09
VDOM cannot be rendered after a component


Archmonger:
### Current Situation

`v0.40.0` brought a regression where VDOM elements cannot be rendered following a component.

For example, `Text 2` is never rendered in the following scenario:

```python
from idom import component, html, run


@component
def Test():
    return html.div("Hello World!")


@component
def App():
    return html.div(
        html.h1("Text 1"),
        Test(),
        html.h1("Text 2"),
    )

run(App)
```

### Proposed Actions

This is a pretty major breaking bug, and needs to be resolved ASAP.

I personally would also yank the `v0.40.0` release as it has been roughly a month of unusably broken behavior.

rmorshea:
I've got an idea on how to bypass the issues with JSON patch. Will try to get to this over the weekend.

rmorshea:
Version has also been yanked.

PR #807, opened 2022-09-11
fix append to stale children ref

rmorshea:
Closes: #806 

A recent change to how components render saved a new list of children to the parent component. In doing so, this caused references to the old list held in other parts of the layout logic to become stale. The solution is to avoid holding onto a reference to a model's children.

## Checklist


Archmonger (approved):
LGTM

Issue #819, opened 2022-10-14
Hook registration framework


Archmonger:
### Current Situation

Currently, every framework needs to import it's hooks (such as `use_websocket`, `use_location`, `use_origin`, etc.) within their own little module.

Would be best to centralize things and have some sort of way to register a hook.

### Proposed Actions

All of the current hooks should be defined a central abstract, and they should be overridden based on the currently active framework. Either that, or those hooks simply become wrappers that call `idom.conf.HOOKS["use_query"]( ... )`

That way, `django-idom` can send its definition of `use_location` to be utilized by IDOM Core.

PR #823, opened 2022-10-19
add use_connection hook

rmorshea:
closes: #819 

Rather than having a hook registration framework. All we need in order to solve the problem at hand is to make it possible for each respective backend to use a common context through which they can declare a scope and location. To do this we define a `ConnectionContext` that each backend must activate with a `Connection` object. The `Connection` object allows the backend to declare scope, location, and an additional, implementation specific, carrier. The carrier is intended to describe the manner in which a connection is mediated (e.g. an HTTP request or websocket).

As a result of these changes, there are now common hooks for the following:

- `use_connection() -> Connection[Any]`
- `use_scope() -> dict[str, Any]`
- `use_location() -> Location`

Backend implementations provide their own `idom.backend.*.use_connection` hooks, but these return the exact same object as the common `idom.use_connection` hook. The only difference is that the type annotation for the return type is narrowed to reflect the carrier of the backend. In narrowing the type though, various backend implementation are also provide implementation specific `use_request` or `use_websocket` hooks.

## Checklist

### DIFF
--- a/tests/test_widgets.py
+++ b/tests/test_widgets.py
@@ -3,6 +3,7 @@ from pathlib import Path
 
 import idom
 from idom.testing import DisplayFixture, poll
+from tests.tooling.common import DEFAULT_TYPE_DELAY
 
 
 HERE = Path(__file__).parent
@@ -86,13 +87,13 @@ async def test_use_linked_inputs(display: DisplayFixture):
     input_1 = await display.page.wait_for_selector("#i_1")
     input_2 = await display.page.wait_for_selector("#i_2")
 
-    await input_1.type("hello", delay=50)
+    await input_1.type("hello", delay=DEFAULT_TYPE_DELAY)
 
     assert (await input_1.evaluate("e => e.value")) == "hello"
     assert (await input_2.evaluate("e => e.value")) == "hello"
 
     await input_2.focus()
-    await input_2.type(" world", delay=50)
+    await input_2.type(" world", delay=DEFAULT_TYPE_DELAY)
 
     assert (await input_1.evaluate("e => e.value")) == "hello world"
     assert (await input_2.evaluate("e => e.value")) == "hello world"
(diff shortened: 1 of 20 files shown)


rmorshea:
@Archmonger to make this work with `django-idom` the approach would be remove the [custom `WebsocketContext`](https://github.com/idom-team/django-idom/blob/main/src/django_idom/websocket/consumer.py#L77) in favor of IDOM core's `ConnectionContext`. See the `starlette` implementation for [an example](https://github.com/idom-team/idom/blob/a35048fa325c98d10673b5c2e327d24c023fa7ff/src/idom/backend/starlette.py#L156-L163).

rmorshea:
@Archmonger does this look good to you?

Archmonger:
Been pretty busy these last two weeks, haven't had time to do a code review. Hopefully will have time at night during the upcoming weekdays.

Archmonger (changes requested):
I think we shouldn't go all-in on changing `websocket` to `carrier`.

The more I think about our reliance on an asyncio running loop, the more impossible it seems for us to use anything other than Websockets.

`HTTP` simply doesn't have any decent level of persistence, and it's awkward/unreasonable to make every user run an "IDOM server" separately as a backhaul running loop. I've been debating closing that issue related to this.

Everything else LGTM.

rmorshea:
I don't really see any harm in using a generic name like `carrier`. Connections could be mediated by websockets, regular sockets, a multiprocessing pipe, or even some in-memory construct depending on the use case. For example, I could see a pipe or in-memory construct being used if we ever developed some integration with a terminal interface like [textual](https://github.com/Textualize/textual). If you actually need access to the carrier you probably know that it's a websocket you're dealing with, in which case you're going to use a backend specific hook (e.g. `use_websocket`) anyway.

tests/test_widgets.py
```diff
@@ -86,13 +86,13 @@ def SomeComponent():
     input_1 = await display.page.wait_for_selector("#i_1")
     input_2 = await display.page.wait_for_selector("#i_2")
 
-    await input_1.type("hello", delay=20)
+    await input_1.type("hello", delay=50)
```
Archmonger, line 89: I recommend setting this to 250ms, and storing it within a global.

rmorshea: Also created an issue to be able to set a default delay https://github.com/microsoft/playwright-python/issues/1614

rmorshea:
I'd like to merge this so I can get started integrating these changes into idom-router. I don't think there's anything that will be impacted if we decide to rename the `carrier` attribute so we can discuss that possibility in an issue if needed.

Issue #796, opened 2022-08-04
Snowpack is Deprecated


rmorshea:
### Current Situation

Snowpack (our build tool for the IDOM client) was recently deprecated

### Proposed Actions

On their site they recommend using [Vite](https://vitejs.dev/) instead.

PR #824, opened 2022-10-20
Vite

rmorshea:
closes: #796

This is low priority, but unfortunately it involves some API changes that I'd like to get out of the way. Currently, the server exposes the following interface:

```
/<prefix>/<path>/_api/stream
/<prefix>/<path>/_api/modules/<file-path>
/<prefix>/<path>/<file-path>
```

This lead to some complicated route patterns. Under these changes though, the server exposes a simpler interface:

```
/_idom/assets/<file-path>
/_idom/stream/<spa-path>
/_idom/modules/<file-path>
/<spa-path>
```

Here, all of IDOM's core server APIs are housed under a `/_idom` root URL. This makes it simpler for the client to direct requests to the correct location and, I assume, makes it possible for the browser to cache static resources. The only complication is that the client needs to coordinate requesting `/_idom/stream/<spa-path>` where `<spa-path>` is the same from `/<spa-path>` where the SPA is being hosted. Backend implementations should then normalize `<spa-path>` when constructing `Connection.location` by stripping any URL prefix.

## Checklist


rmorshea:
@acivitillo, I don't know if you're still using IDOM, but if you are, this may impact you.

Archmonger (approved):
LGTM.

Issue #616, opened 2022-01-27
MyPy plugin to warn for component unpacking operators


rmorshea:
### Current Situation

Currently we don't document that using the unpacking operator (`*args`) has undesirable consequences on whether we do or don't warn about missing keys.

### Proposed Actions

Document this fact. If it turns out that, despite this documentation, people run into this issue often, [consider changing the syntax](https://github.com/idom-team/idom/issues/615).

### Work Items

- [ ] add docs on unpacking operator to "Rendering Data" docs.

Archmonger:
@rmorshea Can I add a `Todo (Docs)` column to the [GitHub Project's page](https://github.com/orgs/idom-team/projects/1/views/5) to prevent documentation tasks from cluttering the task list?

It's hard to tell at first glance how much development actually needs to go into v1 since there's so many docs changes needed.

rmorshea:
Go for it. Let me know if you're not able to.

rmorshea:
I think [a MyPy plugin](https://github.com/idom-team/idom/pull/825) could help with this

Archmonger:
Changing this to issue a plugin, and moving to V2.

Archmonger:
Closing this as I intend to lean us away from `mypy` in near future.

All other reactive-python repos have already been migrated to `pyright` for type checking, which is substantially faster and generates far less annoying/useless errors.

Issue #615, opened 2022-01-25
Warn or prevent using unpack operator within components


Archmonger:
Using the unpack operator is dangerous within a component's `return`. ReactPy relies on the position of child elements to keep track of hooks (for any element that doesn't have a `key=...` defined). We'll either need to warn against using it, or completely prevent it from being used.

There is currently has no way of providing "key does not exist for dynamic children" warnings when using the unpack operator.

## Original Discussion
https://github.com/idom-team/idom/discussions/607

Archmonger:
I still want to note, it is still viable to not develop this and just appropriately document the limitation.

rmorshea:
Ok, here's what I think we should do here:

1. clearly document this.
2. see how commonly people report problems with this.
3. based on the commonality of the problem consider making the syntax change as part of a 2.0 release.

Archmonger:
Closing this in favor of #918, given the following statement provided above:
> React also has a similar issue with the Javascript spread operator if you're not using JSX syntax. Though, everyone uses JSX so it doesn't really come up.

PR #825, opened 2022-10-20
add mypy plugin

rmorshea:
Partially addresses: https://github.com/idom-team/idom/issues/794
closes: #616
closes: #615 

This works around the problem that the `@component` decorator hides the signature of the underlying function. Also check for improperly passing children via `*args`

All you need to do to run this is configure MyPy with:

```
[mypy]
plugins = idom.mypy
```

## Checklist

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -6,6 +6,7 @@ ignore_missing_imports = True
 warn_unused_configs = True
 warn_redundant_casts = True
 warn_unused_ignores = True
+plugins = idom.mypy
 
 [flake8]
 ignore = E203, E266, E501, W503, F811, N802, N806
@@ -41,6 +42,7 @@ exclude_lines =
     raise NotImplementedError
 omit =
     src/idom/__main__.py
+    src/idom/mypy.py
     src/idom/core/_fixed_jsonpatch.py
 
 [build_sphinx]
--- a/src/idom/mypy.py
+++ b/src/idom/mypy.py
@@ -1,4 +1,6 @@
-from typing import Callable, Final
+from __future__ import annotations
+
+from typing import Any, Callable, Final
 
 from mypy.errorcodes import ErrorCode
 from mypy.nodes import ArgKind
@@ -6,9 +8,15 @@ from mypy.plugin import FunctionContext, Plugin
 from mypy.types import CallableType, Instance, Type
 
 
-KEY_IN_RENDER_FUNC: Final = ErrorCode(
+KEY_IN_RENDER_FUNC_ERROR: Final = ErrorCode(
     "idom-key-in-render-func",
     "Component render function has reserved 'key' parameter",
+    "IDOM",
+)
+
+NO_STAR_ARGS_ERROR: Final = ErrorCode(
+    "idom-no-star-args",
+    "Children were passed using *args instead of as a list or tuple",
     "IDOM",
 )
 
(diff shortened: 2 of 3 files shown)


src/idom/mypy.py
```diff
@@ -0,0 +1,51 @@
...
+    "Component render function has reserved 'key' parameter",
+    "IDOM",
+)
+
+
+class MypyPlugin(Plugin):
+    """MyPy plugin for IDOM"""
+
+    def get_function_hook(self, fullname: str) -> Callable[[FunctionContext], Type]:
+        if fullname == "idom.core.component.component":
```
Archmonger, line 20: Are we allowed to do an `from idom import component` and then have this check be `fullname == f"{component.__module}.{component.__name__}"`? 

Would improve refactorability.

setup.cfg
```diff
@@ -6,6 +6,7 @@ ignore_missing_imports = True
 warn_unused_configs = True
 warn_redundant_casts = True
 warn_unused_ignores = True
+plugins = idom.mypy
```
Archmonger, line 9: We need to document this somehow.

Undocumented features essentially do not exist.

Archmonger (changes requested):
This is not the ideal solution, as it prioritizes MyPy over any other type checkers (ex `pylance`).

We will need to create an issue to revisit this when Python 3.11 becomes mainstream with better type hinting support

src/idom/mypy.py
Archmonger: Yes, however, if we decide to change the path `ido.core.component.component` in the future, this plugin will break silently.

I would recommend using python imports to generate that path, that way refactoring libraries such as [`rope`](https://github.com/python-rope/rope) can automatically update these paths. Or at the very least, this plugin would create a visible exception to demonstrate that things are broken.

rmorshea:
> This is not the ideal solution, as it prioritizes MyPy over any other type checkers (ex pylance).

I realize that, but there simply isn't any way to annotate what `@component` does with the typing tools Python provides right now. PyRight, the type checker in PyLance, also has [no support for plugins](https://github.com/microsoft/pyright/issues/607) and it doesn't seem like there will be any time soon. 

Other very large projects have already opted to write MyPy plugins. For example, [Numpy](https://numpy.org/devdocs/reference/typing.html#mypy-plugin) and [Pydantic](https://pydantic-docs.helpmanual.io/mypy_plugin/). So I don't think it's too crazy for us to do the same.

> We will need to create an issue to revisit this when Python 3.11 becomes mainstream

Typically, even if typing tools are included in newer versions of Python, they can still be used through [`typing_extensions`](https://github.com/python/typing_extensions). This is how we use newer features like `Protocol` and `ParamSpec` even though those were not released in Python 3.7 (the oldest version of Python we support). But again, there are no typing PEPs that have been accepted thus far that would allow us to appropriately annotate what `@component` does, especially after we make [this change](https://github.com/idom-team/idom/issues/755).

> Also, test seem broken for potentially unrelated reasons.

This is addressed in https://github.com/idom-team/idom/pull/823

temp.py
```diff
@@ -0,0 +1,9 @@
+from idom import component
+
+
+@component
+def temp(x: int, y: int) -> None:
+    return None
+
+
+temp(*[1, 2])
```
Archmonger, line 9: We might want to put `temp.py` into the gitignore if you frequently use it for testing

Archmonger:
I have been experimenting with the new [OpenAI Public Beta](https://chat.openai.com/chat) to see if it can auto complete this PR.

## Input
Create a MyPy plugin that only warns if the unpacking operator (`*` or `**`) is used within an `idom.html.*` function call of within a decorated function. This must only apply to functions decorated with `@component`. The `@component` decorator path must resolve to `idom.core.component.component`

## Output

Here is a possible implementation of a MyPy plugin that satisfies the requirements you specified:

```python
import ast
from typing import Any, Dict, List, Tuple

from mypy.plugin import FunctionContext, MethodContext, Plugin
from mypy.types import CallableType

def is_idom_html_function(func_name: str) -> bool:
    return func_name.startswith('idom.html.')

def is_component_decorated(decorators: List[ast.AST]) -> bool:
    if not decorators:
        return False
    return any(
        isinstance(decorator, ast.Name) and decorator.id == 'component'
        for decorator in decorators
    )

class IdomMypyPlugin(Plugin):
    def get_function_hook(self, fullname: str) -> Optional[Callable[[FunctionContext], Type]]:
        if fullname == 'idom.core.component.component':
            return function_hook

    def function_hook(ctx: FunctionContext) -> Type:
        sig = ctx.default_return_type
        if not isinstance(sig, CallableType):
            return sig
        if not is_component_decorated(ctx.func.decorators):
            return sig
        if not sig.is_var_arg:
            return sig
        ctx.api.warn_unused('W001', ctx.context, ctx.func, 'Unpacking operator used in decorated function')
        return sig
```

This plugin defines a single function hook that is called for any function with the full name idom.core.component.component. This function checks if the function being called is decorated with @component, and if it is, it checks if the function has a variable argument (*args or **kwargs) and if it does, it emits a warning using the MyPy API.

Note that this implementation is just an example and may not be complete or correct in all cases. MyPy plugins can be complex to write, so it's important to thoroughly test and debug them before using them in production.

rmorshea:
Gonna close this. We can return to this for reference when the time comes.

PR #832, opened 2022-11-09
`html_to_vdom` transform to remove html/body but preserve head content

Archmonger:
There's currently some awkwardness in converting `django` views to VDOM, due to the fact that the head content is never loaded into the page.

For example, CSS/JS that is within `<head>` is simply never added to the page.

This PR makes `html_to_vdom` use the same approach browsers use when merging two HTML/body node trees. Namely, the contents of the `<head>` and `<body>` tags are directly spit out onto the page, while removing those top level tags themselves.

## Checklist

- [X] GitHub Issues which may be closed by this Pull Request have been linked.


rmorshea:
Could this be implemented as a transform instead?

rmorshea:
I don't think we should bake this into `html_to_vdom`. It's not clear to me that this approach to dealing with html/head/body is universally applicable. For example, in the future, we could allow users to declare head elements in IDOM by leveraging a tool like [react-helmet](https://github.com/nfl/react-helmet). In which case, we'd actually want to preserve the head/body distinction. Thus I think this should be implemented as a transform. One that could be included in IDOM core given its broad, but not necessarily universal, appeal.

Archmonger:
Given that helmet exists, I'm okay with that. On a different note, is there truly a purpose for us to keep the `idom.html.html` and `idom.html.body` representations?

rmorshea:
We could, though there's not much cost to having them there either. If you feel strongly about it, I'd be ok removing them.

Archmonger:
It has been converted to a transform.

Archmonger:
Let me know if deleting `body` is something you want to commit to. I personally think removing it helps set expectations, since currently I don't see any way to use it.

rmorshea (dismissed):

src/idom/utils.py
```diff
@@ -223,3 +208,40 @@ def _hypen_to_camel_case(string: str) -> str:
 
 class HTMLParseError(etree.LxmlSyntaxError):  # type: ignore[misc]
     """Raised when an HTML document cannot be parsed using strict parsing."""
+
+
+def del_html_body_transform(vdom: dict[str, Any]) -> dict[str, Any]:
```
rmorshea, line 213:
```suggestion
def del_html_body_transform(vdom: VdomDict) -> VdomDict:
```

Archmonger:
I didn't have it set to VdomDict because mypy didn't like that for some reason. Was easier to set as a generic dict than figure it out at the time.

Archmonger:
I reverted the `VdomDict` type hint change since I wasn't entirely sure what was wrong.

rmorshea:
Could this transform be more simply written as?

```python
def del_html_body_transform(vdom: VdomDict) -> VdomDict:
    """Removes ``<html>``, ``<head>``, and ``<body>`` elements.

    This is accomplished by turning any matching elements into fragments.
    See :func:`idom.html._` for more details on element fragments.

    .. note::

        Intended for use with :func:`html_to_vdom`.

    Parameters:
        vdom:
            The VDOM dictionary to transform.
    """
    return {**vdom, "tagName": ""} if vdom["tagName"] in ("html", "body", "head") else vdom
```

Archmonger:
Yeah.

Results in an uglier final VDOM with a bunch of fragments, but functionally would work the same when rendered.

src/idom/utils.py
```diff
@@ -289,6 +272,20 @@ def _hypen_to_camel_case(string: str) -> str:
     return first.lower() + remainder.title().replace("-", "")
 
 
+def del_html_head_body_transform(vdom: dict[str, Any]) -> dict[str, Any]:
```
rmorshea, line 275: Annotating as `VdomDict` should work now.

```suggestion
def del_html_head_body_transform(vdom: VdomDict) -> VdomDict:
```

Issue #834, opened 2022-11-14
Allow users to configure client <head/>


rmorshea:
### Current Situation

At the moment, users are unable to configure the `<head/>` element of IDOM's client. They can use a `script()` to change details of the page after the fact, but this does not help in the case the user needs finer control of how/when resources load or for SEO.

Related discussions: #485, #833

### Proposed Actions

Add a `head` parameter to all the built-in backends that accepts an HTML string or `VdomDict`. We'll need https://github.com/idom-team/idom/issues/435 to accomplish this.

Issue #435, opened 2021-07-17
add VDOM to HTML util function


rmorshea:
Sample implementation:

```python
from html import escape as html_escape

def vdom_to_html(vdom: Union[str, VdomDict]) -> str:
    """Convert a VDOM dictionary into an HTML string

    Only the following keys are translated to HTML:

    - ``tagName``
    - ``attributes``
    - ``children`` (must be strings or more VDOM dicts)
    """
    if isinstance(vdom, str):
        return vdom

    try:
        tag = vdom["tagName"]
    except TypeError as error:
        raise TypeError(f"Expected a VDOM dictionary or string, not {vdom}") from error

    if "attributes" in vdom:
        attributes = " " + " ".join(
            f'{k}="{html_escape(v)}"' for k, v in vdom["attributes"].items()
        )
    else:
        attributes = ""

    if "children" in vdom:
        children = "".join(map(vdom_to_html, vdom["children"]))
    else:
        children = ""

    return f"<{tag}{attributes}>{children}</{tag}>"
```

rmorshea:
@Archmonger there's the function.

Archmonger:
I believe I was looking for the inverse of this. Would be useful for Django purposes (ex. render via Django template, push template into IDOM).

rmorshea:
That actually exists 🙂: [`html_to_vdom`](http://idom-docs.herokuapp.com/docs/auto/api-reference.html?highlight=html_to_vdom#idom.utils.html_to_vdom)

Archmonger:
Yep, but if I recall the conservation was brought up in reference to it not being documented.

rmorshea:
The documentation is linked to above. Maybe I'm misunderstanding?

Archmonger:
Oh that's my bad. I must have missed it when scrolling through the docs.

rmorshea:
@acivitillo, if you want to use IDOM for HTML templating, this might be a good thing to try contributing.

PR #835, opened 2022-11-14
allow users to configure html head

rmorshea:
closes: #834
closes: #435

## Checklist


src/idom/utils.py
```diff
@@ -56,6 +62,61 @@ def __repr__(self) -> str:
...
+
+    - ``tagName``
+    - ``attributes``
+    - ``children`` (must be strings or more VDOM dicts)
+    """
+    if isinstance(vdom, str):
+        return vdom
+
+    try:
+        tag = vdom["tagName"]
```
Archmonger, line 78: Don't forget to add a conditional to handle react fragments.

src/idom/utils.py
```diff
@@ -56,6 +62,92 @@ def __repr__(self) -> str:
...
+def vdom_to_html(value: str | VdomDict, indent: int = 0, depth: int = 0) -> str:
+    """Convert a VDOM dictionary into an HTML string
+
+    Only the following keys are translated to HTML:
+
+    - ``tagName``
+    - ``attributes``
+    - ``children`` (must be strings or more VDOM dicts)
+    """
+    if indent:
```
Archmonger, line 74: I don't see a major reason to do indentation within this utility.

If the user really wanted things looking pretty, it's best done via a HTML beautifier like `html5print` or `Prettier`, which would give them way more configuration options.

src/idom/utils.py
```diff
@@ -56,6 +62,92 @@ def __repr__(self) -> str:
...
+        children = "".join(children_list)
+
+    else:
+        children = ""
+
+    if not children:
+        return f"{open_indent}<{tag}{attributes} />" if tag else ""
+    else:
+        return (
+            f"{open_indent}<{tag}{attributes}>{children}{close_indent}</{tag}>"
```
Archmonger, line 133: The current implementation of `vdom_to_html` is currently a bit complex.

It might be cleaner and potentially less error prone to convert from

VDOM -> `lxml.etree` -> `etree.tostring`

rmorshea: I think things will be simplified if I remove the indent option. If not I'll try that out.

rmorshea: Much of the complication came from warning about, and then excluding, things we don't strictly know how to convert to HTML. In the latest changes, I simply cast such values to strings and HTML escape them.

src/idom/utils.py
```diff
@@ -56,6 +62,52 @@ def __repr__(self) -> str:
...
+        for c in value.get("children", ())
+    )
+
+    return (
+        (
+            f"<{tag}{attributes}>{children}</{tag}>"
+            if children
+            # To be safe we mark elements without children as self-closing.
+            # https://html.spec.whatwg.org/multipage/syntax.html#foreign-elements
+            else (f"<{tag}{attributes} />" if attributes else f"<{tag}/>")
```
Archmonger, line 104: This does not properly handle situations like `` tags.

Can potentially considering `lxml` to handle this automatically.

rmorshea: So there's some value in `lxml` for this, but the way LXML handles text turns out to make things more complicated on some ways.

src/idom/utils.py
```diff
@@ -221,5 +272,36 @@ def _hypen_to_camel_case(string: str) -> str:
...
+
+# see list of HTML attributes with dashes in them:
+# https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes#attribute_list
+_CAMEL_TO_DASH_CASE_HTML_ATTRS = {
+    "acceptCharset": "accept-charset",
+    "httpEquiv": "http-equiv",
+}
+
+
+def _vdom_to_html_attr(key: str, value: Any) -> tuple[str, str]:
```
Archmonger, line 286: `_vdom_attr_to_html` sounds like a more appropriate name

src/idom/utils.py
```diff
@@ -221,5 +272,36 @@ def _hypen_to_camel_case(string: str) -> str:
...
+                # We lower only to normalize - CSS is case-insensitive:
+                # https://www.w3.org/TR/css-fonts-3/#font-family-casing
+                f"{_CAMEL_CASE_SUB_PATTERN.sub('-', k).lower()}:{v}"
+                for k, v in value.items()
+            )
+    elif key.startswith("data"):
+        # Handle data-* attribute names
+        key = _CAMEL_CASE_SUB_PATTERN.sub("-", key)
+    else:
+        key = _CAMEL_TO_DASH_CASE_HTML_ATTRS.get(key, key)
```
Archmonger, line 299: Would be more future proof to convert all attributes be converted from `camelCase` to `kebab-case`, rather than just `acceptCharset` and `httpEquiv`.

src/idom/utils.py
```diff
@@ -56,6 +59,24 @@ def __repr__(self) -> str:
...
+    Only the following keys are translated to HTML:
+
+    - ``tagName``
+    - ``attributes``
+    - ``children`` (must be strings or more VDOM dicts)
+    """
+    temp_root = etree.Element("__temp__")
+    _add_vdom_to_etree(temp_root, value)
+    return (
+        cast(bytes, tostring(temp_root)).decode()
```
Archmonger, line 74: The way this return is written makes it harder to read. We should use some variables for improved clarity.

```python
html = cast(bytes, tostring(temp_root)).decode()

# strip out temp root <__temp__> element
return html[10:-11]
```

src/idom/utils.py
```diff
@@ -56,6 +59,24 @@ def __repr__(self) -> str:
         return f"{type(self).__name__}({current})"
 
 
+def vdom_to_html(value: VdomDict) -> str:
```
Archmonger, line 62: change `value` to `vdom`

src/idom/utils.py
```diff
@@ -56,6 +59,24 @@ def __repr__(self) -> str:
...
+    """Convert a VDOM dictionary into an HTML string
+
+    Only the following keys are translated to HTML:
+
+    - ``tagName``
+    - ``attributes``
+    - ``children`` (must be strings or more VDOM dicts)
+    """
```
Archmonger, line 70: Docstring missing parameters

src/idom/utils.py
```diff
@@ -165,6 +190,32 @@ def _etree_to_vdom(
...
+        element = parent
+
+    for c in vdom.get("children", []):
+        if isinstance(c, dict):
+            _add_vdom_to_etree(element, c)
+        elif len(element):
+            last_child = element[-1]
+            last_child.tail = f"{last_child.tail or ''}{c}"
+        else:
+            element.text = f"{element.text or ''}{c}"
```
Archmonger, line 216: These conditionals could use some comments.

Archmonger (changes requested):
Overall looks pretty good, mostly minor nitpicks

src/idom/utils.py
rmorshea: The trouble is that for cases like `tabIndex`, the conversion is `tabindex`, not `tab-index`. Not really sure there's a great alternative.

rmorshea: I can re-writing it, but the current implementation only converts if the key is present in the mapping.

Issue #755, opened 2022-05-23
Unify Component and VDOM Element Constructor Interface


rmorshea:
### Current Situation

Currently the `vdom` element constructor and component constructors have different interfaces. Component constructors are effectively just the same as the functions which define them. For the `vdom` constructor though we pass `vdom(tag, attrs_dict, *children)`. The fact that these two constructors have different interfaces won't play well with this [tagged strings PEP](https://github.com/jimbaker/tagstr/issues/9).

### Proposed Actions

In anticipation of the work from the [tagged strings PEP](https://github.com/jimbaker/tagstr) we should have a unified interface for constructing elements and components. I think the best way to do this would be to effectively copy the interface for React's functional components.

When constructing a component or element the interface would be:

```python
constructor(*children, key=..., **attributes)
```

However the definition for the constructor would be:

```python
def constructor(children: list[any], **attributes: any): ...
```

In the function definition, `*children` gets bundled together into a special `children` argument that gets passed to the function. Component functions without a `children` argument then do not support `*children`. Additionally (as is currently the case) the special `key` parameter is hidden and handled separately.

One wrinkle in this is that the attribute for most user defined components will be snake_case while the attributes for standard HTML elements are camelCase. I suspect this will be confusing for most users. As a result, I suggest that, for standard HTML elements, we include a snake_case to camelCase conversion. Thus something like `div(background_color)` would be treated the same as `div(backgroundColor=...)`.

This will be a relatively large change, so we will provide a way for users to transition to the new interface.

#### Notes

We should probably do https://github.com/idom-team/idom/issues/545 at the same time. So we'd add a `fallback` kwarg to the constructor:

```python
constructor(*children, key=..., fallback=..., **attributes)
```

rmorshea:
@Archmonger thoughts on this?

Archmonger:
I would recommend automatically changing all attributes from snake_case to camelCase on the backend.

I vaguely recall seeing some regex for this in Django core.

rmorshea:
I think I want to implement the conversion client-side since that's where they have to be interpreted, but the effect for the user would be the same - all attributes would automatically become camelCase.

Archmonger:
looks like there's a couple good solutions to this on the JS end as well

https://stackoverflow.com/questions/6660977/convert-hyphens-to-camel-case-camelcase
https://stackoverflow.com/questions/40710628/how-to-convert-snake-case-to-camelcase-in-my-app

rmorshea:
[...]

PR #841, opened 2022-11-29
Unify vdom interface

rmorshea:
closes: #755

## Checklist


Archmonger:
I did some more thinking, `id` would be the the single property that makes this implementation messy. We don't want to override the built-in python `id` function, so do we prefix it with a `_`?

Archmonger:
Discussed this via call with Ryan.

All suffixed underscores should be stripped out.

Archmonger:
We might want to have another discussion as to whether we should move forward with this.

Given both options have positives/negatives, I'm still undecided.

---

## Using Dict
### Pros
- Feels more like ReactJS props (since in React/JSX props are "at the top" within each HTML tag)
- Does not have an issue with Python keywords`

### Cons
- Might feel weird to some
- Currently requires using `camelCase` within the dict, which is not pythonic.

## Using Kwargs
### Pros
- Feels more pythonic
- Can do long-form type hints for all available properties

### Cons
- Does not have good visual similarity to ReactJS
- Switching to this will require developer migration efforts
- Will feel weird to use for `data-*` and `aria-*` attributes
- Makes writing conditional parameters more awkward, since they'll need to be an unpacked dict.

rmorshea:
To clarify, only the following are [keywords](https://docs.python.org/3/reference/lexical_analysis.html#keywords) would require you to add a trailing `_`:

```
False      await      else       import     pass
None       break      except     in         raise
True       class      finally    is         return
and        continue   for        lambda     try
as         def        from       nonlocal   while
assert     del        global     not        with
async      elif       if         or         yield
```

Functions like `id`, `map`, or `type` which are available in the global namespace, but are not keywords do not conflict.

Archmonger:
Do we want the conversion script to be able to handle props in a `dict` variable?

```python
MODAL_CONTAINER = {
    "id": "modal-container",
    "className": "modal fade",
    "tabIndex": "-1",
    "aria-hidden": "true",
}

@component
def modal():
    return div(
        MODAL_CONTAINER,
        ....
    )

**Comment by rmorshea** [comment]
Unfortunately, handling that is a lot more complicated. Looking at the `div` call in isolation, there's no way to tell if it's a child element or an attribute dictionary. You'd have to work backwards and determine the type of the variable. It's probably possible, but a lot more challenging.

**Comment by rmorshea** [comment]
@Archmonger can you try running [the script](https://github.com/idom-team/idom/blob/e8231ff45eacf94d860a45aa33a065d90201715c/scripts/fix_vdom_constructor_usage.py) against some of your code? It fairs pretty well against IDOM's source but that's a fairly small sample size. You should be able to copy and paste in an environment where IDOM is installed and do the following:

```
python fix_vdom_constructor_usage.py [patterns]
```

Where patterns as any number of glob pattern for the files you want to modify for example:

```python
python fix_vdom_constructor_usage.py tests/**/*.py tests/*.py
```

Archmonger:
Running the output on `conreq@app_store` then did a visual inspection of all the deltas.

It only messed in in one situation, where it forgot to leave a preceding space on an inline `... if .... else ...` statement [on this LOC](https://github.com/Archmonger/Conreq/blob/9c6e80d002f91b8609bcd4d3cfcfa36f389c3a83/conreq/_core/sign_up/components.py#L22)

rmorshea:
Outside of the updating narrative docs this change is pretty much complete. Usages of `vdom()` where a child is a dictionary without a `tagName` key produce a warning explaining the API change. IDOM will also distribute a CLI utility for upgrading to the new API (e.x. `idom update-html-usages [PATTERNS]`).

Archmonger:
We probably want that CLI to be a bit more user friendly.

`idom update-html-usages [DIR_OR_FILE]`

If we inspect the path to be a dir, then we should automatically create the glob `**/*.py`

Otherwise, only format one file.

rmorshea:
Yup. Just made that change.

Archmonger:
Will review when tests pass

rmorshea:
There's some minor kinks in the tests to work out, but this is ready for review.

docs/examples.py
```diff
@@ -122,7 +122,7 @@ def Wrapper():
     def PrintView():
         text, set_text = idom.hooks.use_state(print_buffer.getvalue())
         print_buffer.set_callback(set_text)
-        return idom.html.pre({"class": "printout"}, text) if text else idom.html.div()
+        return idom.html.pre(text, class_="printout") if text else idom.html.div()
```
Archmonger: Should this be `class_name`?

src/idom/utils.py
```diff
@@ -233,9 +232,9 @@ def _mutate_vdom(vdom: VdomDict) -> None:
         # Convince type checker that it's safe to mutate attributes
         assert isinstance(vdom["attributes"], dict)
 
-        # Convert style attribute from str -> dict with camelCase keys
+        # Convert style attribute from str -> dict with snake case keys
         vdom["attributes"]["style"] = {
-            _hypen_to_camel_case(key.strip()): value.strip()
+            key.strip().replace("-", "_"): value.strip()
```
Archmonger, line 237: I think we should assume that all dashes (`-`) are intentional.

Archmonger (changes requested):
Not many nitpicks on this one.

Can we potentially sneak in a revert to the [use_state type hints](https://github.com/idom-team/idom/issues/794#issuecomment-1398418203) on this PR? It's fairly painful to to `use_state` right now.

src/idom/_console/update_html_usages.py
```diff
@@ -0,0 +1,231 @@
...
+
+    Additionally, We are unable to perserve the location of comments that lie within any
+    rewritten code. This command will place the comments in the code it plans to rewrite
+    just above its changes. As such it requires manual intervention to put those
+    comments back in their original location.
+    """
+    if sys.version_info < (3, 9):  # pragma: no cover
+        raise RuntimeError("This command requires Python>=3.9")
+    for d in directories:
+        for file in Path(d).rglob("*.py"):
```
Archmonger: Does this successfully get all subdirectories as well?

rmorshea: Yes `rglob` as apposed to `glob` is recursive.

Archmonger:
A general question remains: do you think this solution is portable enough to be used conveniently with `custom backends`?

Would this solution be more portable if introduced into the JavaScript client?

I think the answer is the current implementation is good, but I thought it's worthwhile to raise the question now.

docs/examples.py
rmorshea: Yeah, I think this is actually a typo in the original example.

docs/source/reference/_examples/simple_dashboard.py
```diff
@@ -83,10 +83,10 @@ def update_value(value):
         set_value_callback(value)
 
     return idom.html.fieldset(
-        {"class": "number-input-container"},
-        idom.html.legend({"style": {"font-size": "medium"}}, label),
+        idom.html.legend(label, style={"font_size": "medium"}),
         Input(update_value, "number", value, attributes=attrs, cast=float),
         Input(update_value, "range", value, attributes=attrs, cast=float),
+        class_="number-input-container",
```
Archmonger: Another situation that should be `class_name`

src/idom/utils.py
rmorshea: We actually need to do this. Previously we converted directly to camelCase because this is what React expects client-side. Now though, purely for the sake of consistency we convert to snake_case, which the client subsequently converts to camelCase.

rmorshea:
> do you think this solution is portable enough to be used conveniently with custom backends?

It's a little unclear what you're asking. What part of this change are you thinking might not be portable between backends?

src/idom/utils.py
Archmonger: Perhaps I'm not understanding, but I don't see any reason why we couldn't directly communicate props with dashes (`-`) to React.

Seems to be [a valid prop value](https://stackoverflow.com/questions/27285856/how-do-i-dynamically-set-html5-data-attributes-using-react)?

Not a big deal though, since it's just `data-` and `aria-` attributes.

Archmonger:
I suppose the question is, if someone was to implement a custom backend for a new Python framework, would they need to worry about any additional server-side key mutations as a result of this PR?

And if so, should all those mutations occur on the client side to alleviate this?

---

As I mentioned, based on what I saw on this PR I think the answer is the current implementation is portable enough to be used with `Custom Backends` without any additional work. But I thought it'd be worth double checking.

src/idom/utils.py
rmorshea: This conversion happens to only impact the `style` dict however, in general, we [handle data/aria attributes](https://github.com/idom-team/idom/blob/ddc7178c7bbf8efcedac5c6ac43308eb059504bd/src/client/packages/idom-client-react/src/element-utils.js#L54-L78) in the client. Specifically, the behavior is, if your attribute name begins with `data_` or `aria_`, we convert all `_` to `-` characters.

rmorshea:
In general backends are fairly agnostic to the way and format in which layouts are rendered. This is somewhat indicated by the fact that this change doesn't really touch any files in `idom/backend`. With that said, backends could be further insulated by implementing a message routing system. In doing so any changes in the things consuming messages wouldn't impact the backend, since all the backend would be concerned with is sending websocket messages to/from the routing system.

Archmonger (dismissed):

Archmonger:
When we cut a 1.0.0 release of IDOM, I will sneak it into the 3.0.0 release of Django IDOM

Archmonger:
On a different note, in the future releases of IDOM, I recommend we switch to [semantic versioning](https://semver.org/spec/v2.0.0.html) rather than feature-set versioning.

We will need to rename our GitHub Projects milestones as such.

rmorshea:
Completely forgot to update docs/changelog will create a followup PR

Issue #574, opened 2022-01-10
Serialize File Content


rmorshea:
Right now we don't serialize file content. Just metadata. Along with this we'll want to allow users to configure the max message size to help users protect against large file upload attacks.

See [this comment](https://github.com/reactive-python/reactpy/issues/574#issuecomment-1595498507) for implementation details.

Archmonger:
Just had a thought.

We could possibly leave file serialization to HTTP, and handle things on the backend.

For example, adding a HTTP endpoint

```python

# Uploads are given a unique UUID path to route to
path("_idom/upload/{uuid}")
```

Not really sure how we'd be able to tell the WS side of things when an upload is a complete though...

Archmonger:
If we do go down this road of serializing via websockets, it's going to be extremely important to stream the file rather than buffer it in memory. Otherwise, this would be a DOS attack vector.

Ideally, I'd like file stuff to be handled by individual frameworks... but the design of that might not be feasible.

Archmonger:
Something else to keep in mind if we serialize over WS:

We should allow data to be send in an arbitrary order of packets. Which means while sending things in chunks, we'll need to prepend some header values, such as a `data-addr`, which signifies what "memory location" in the file that chunk is related to.

Archmonger:
[PyZMQ](https://github.com/zeromq/pyzmq) might be a decent option to communicate between separate ASGI processes.

rmorshea:
The initial implementation should probably use form elements to upload files over HTTP. We can overwrite the default `onSubmit` behavior for forms to submit without a redirect. Unfortunately though, I don't think we can supply a built-in solution that will allow for inter-process communication - that will have to be supplied on a case-by-case basis depending on the user's deployment.

With that said, I have some thoughts on how this could be accomplished via the websocket in a more natural way, but it would require some pretty deep changes to how we send messages between the client and the server. Right now, there is no real messaging protocol - the server always sends layout updates and the client always sends layout events. We'd need to change this in order to allow the client and server to send different types of messages to each other. For example, all messages between the client and server would probably take the form:

```python
{
  "type": str,  # some unique name
  "version": str,  # allow client and server to sort out compatibility
  ...  # any other data
}
```

Archmonger:
We are likely going to develop [two file upload handlers](https://docs.djangoproject.com/en/4.2/topics/http/file-uploads/#upload-handlers).

For small files, we stream the file directly to memory.
[...]

PR #859, opened 2022-12-29
use fetch request to upload form data

rmorshea:
closes: #574 

This change prevents the default behavior of a `<form>` element's `onSubmit` event handler. Instead of navigating away from the current page to submit form data, it is sent to the server via a Javascript `fetch` request. We do not push form data into the URL (this behavior will be left to third party packages like [idom-router](https://github.com/idom-team/idom-router)).

Improving form submission behavior will allow users to use form elements to send, among other things, file data to servers. For example, the following will now be possible:

```python
from idom import component, html
from idom.backend.starlette import configure
from starlette.applications import Starlette

app = Starlette()

async def handle_file_upload(request):
    form = await request.form()
    ...  # do stuff with the form

app.add_route("/file-upload", handle_file_upload, methods=["POST"])

@component
def Example():
    return html.form(
        {
            "enctype": "multipart/form-data",
            "action": "/file-upload",
            "method": "POST",
        },
        html.input({"type": "file", "name": "file", "id": "file-input"}),
        html.input({"type": "submit", "id": "form-submit"}),
    )

configure(app, Example)
```

## Checklist

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -181,6 +181,7 @@ def test_python_suite(session: Session) -> None:
     install_requirements_file(session, "test-env")
     session.run("playwright", "install", "chromium")
     posargs = session.posargs
+    posargs += ["--reruns", "3", "--reruns-delay", "1"]
 
     if "--no-cov" in session.posargs:
         session.log("Coverage won't be checked")
--- a/noxfile.py
+++ b/noxfile.py
@@ -181,7 +181,6 @@ def test_python_suite(session: Session) -> None:
     install_requirements_file(session, "test-env")
     session.run("playwright", "install", "chromium")
     posargs = session.posargs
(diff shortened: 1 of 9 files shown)


rmorshea:
cc: @Archmonger thoughts on this?

Archmonger:
Is there any way to automate this route handling with some combination of `Context` scoping and `uuid` session uniqueness? 

Ideally these route would get created within `from idom.backend.my_chosen_backend import configure`.

See [here](https://github.com/idom-team/idom/issues/574#issuecomment-1087133473)

rmorshea:
After thinking further, I think modifying the default behavior of forms might not have been the best idea. Doing so should probably be the purview of third-party packages (e.g. idom-router a la [React Router's forms](https://reactrouter.com/en/main/components/form)).

Issue #847, opened 2022-12-01
Detached model state on render of context consumer if unmounted and context value does not change.


Archmonger:
### Current Situation

`Context` should be able to be scoped to conditionally rendered components, and when those subcomponents are removed the deletion of the `Context` should be handled gracefully.

See: https://github.com/Archmonger/Conreq/blob/d64b3ba879d263b36ab4806d4790d44cefea1045/conreq/_core/components.py#L46-L64

```
Traceback (most recent call last):
  File "C:\Users\Markg\Documents\Repositories\Conreq\.venv\lib\site-packages\django\contrib\staticfiles\handlers.py", line 101, in __call__
    return await self.application(scope, receive, send)
  File "C:\Users\Markg\Documents\Repositories\Conreq\.venv\lib\site-packages\channels\routing.py", line 62, in __call__
    return await application(scope, receive, send)
  File "C:\Users\Markg\Documents\Repositories\Conreq\.venv\lib\site-packages\channels\security\websocket.py", line 37, in __call__
    return await self.application(scope, receive, send)
  File "C:\Users\Markg\Documents\Repositories\Conreq\.venv\lib\site-packages\channels\sessions.py", line 47, in __call__
    return await self.inner(dict(scope, cookies=cookies), receive, send)
  File "C:\Users\Markg\Documents\Repositories\Conreq\.venv\lib\site-packages\channels\sessions.py", line 263, in __call__
    return await self.inner(wrapper.scope, receive, wrapper.send)
  File "C:\Users\Markg\Documents\Repositories\Conreq\.venv\lib\site-packages\channels\sessions.py", line 47, in __call__
    return await self.inner(dict(scope, cookies=cookies), receive, send)
  File "C:\Users\Markg\Documents\Repositories\Conreq\.venv\lib\site-packages\channels\sessions.py", line 263, in __call__
    return await self.inner(wrapper.scope, receive, wrapper.send)
  File "C:\Users\Markg\Documents\Repositories\Conreq\.venv\lib\site-packages\channels\auth.py", line 185, in __call__
    return await super().__call__(scope, receive, send)
  File "C:\Users\Markg\Documents\Repositories\Conreq\.venv\lib\site-packages\channels\middleware.py", line 24, in __call__
    return await self.inner(scope, receive, send)
  File "C:\Users\Markg\Documents\Repositories\Conreq\.venv\lib\site-packages\channels\routing.py", line 116, in __call__
    return await application(
[...]

PR #870, opened 2023-01-07
remove should_render

### LINKED ISSUE #846: Nested context does no update value if outer context should not render
### Current Situation

ReactJS appears to support nested context.

However, in IDOM situations like the following simply don't render...

```python
    return ModalStateContext(
        HomepageStateContext(
            navbar(),
            modal(),
            sidebar(),
            viewport_loading_animation(),
            viewport(),
            backdrop(),
            value=home_state,
        ),
        value=modal_state,
    )
```

### Proposed Actions

Modify the `ContextProvider` to support nested contexts.

rmorshea:
Closes #846, #847 

## Checklist

### DIFF
--- a/docs/source/about/changelog.rst
+++ b/docs/source/about/changelog.rst
@@ -23,7 +23,18 @@ more info, see the :ref:`Contributor Guide <Creating a Changelog Entry>`.
 Unreleased
 ----------
 
-No changes.
+**Removed**
+
+- :pull:`870` - ``ComponentType.should_render()``. This method was implemented based on
+  reading the React/Preact source code. As it turns out though it seems like it's mostly
+  a vestige from the fact that both these libraries still support class-based
+  components. The ability for components to not render also caused several bugs.
+
+**Fixed**
+
+- :issue:`846` - Nested context does no update value if outer context should not render.
+- :issue:`847` - Detached model state on render of context consumer if unmounted and
+  context value does not change.
 
 
 v0.42.0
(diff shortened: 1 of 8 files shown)


tests/test_core/test_layout.py
```diff
@@ -1238,3 +1160,43 @@ def Child():
                 }
             ],
         }
+
+
+async def test_render_removed_context_consumer():
```
rmorshea, line 1165: @Archmonger if you're curious the code below reproduces that detached model error

rmorshea:
I'm actually ok merging this without review.

tests/test_core/test_layout.py
Archmonger: Is it the nonlocals that are causing the issue?

docs/source/about/changelog.rst
```diff
@@ -23,7 +23,18 @@ more info, see the :ref:`Contributor Guide <Creating a Changelog Entry>`.
...
+**Removed**
+
+- :pull:`870` - ``ComponentType.should_render()``. This method was implemented based on
+  reading the React/Preact source code. As it turns out though it seems like it's mostly
+  a vestige from the fact that both these libraries still support class-based
+  components. The ability for components to not render also caused several bugs.
+
+**Fixed**
+
+- :issue:`846` - Nested context does no update value if outer context should not render.
```
Archmonger, line 35: does not*

tests/test_core/test_layout.py
rmorshea: No, those are just a trick to get a handle on hook callback from outside the components. The comments in the test body hopefully explain the cause. If they aren't clear I can make a follow up PR to improve the description.

Issue #877, opened 2023-01-12
Remove JSON Patch


rmorshea:
closes: #498

### Current Situation

Currently, we've effectively [turned off diffing with JsonPatch](https://github.com/idom-team/idom/pull/809) due to [bugs](https://github.com/stefankoegl/python-json-patch/issues/138) in [python-json-patch](https://github.com/stefankoegl/python-json-patch) which unfortunately does not seem to be actively maintained.

### Proposed Actions

Take the final step of removing JSON patch entirely. Downstream clients will need to be updated accordingly.

rmorshea:
@Archmonger, at the moment, this change proposes that all messages between the server and client take the form:

```python
{
    "header": {
        "type": str,
        "version": str,
    }
}
```

Instead of adding this to each message, the client and server could be expected to perform a handshake in their first two messages in which they declare what message types and versions they use. From that point onward all messages would simply have a single `type` key where all message metadata is already known to the client/server via that initial handshake.

PR #881, opened 2023-01-20
remove json patch

rmorshea:
closes: #877

## Checklist


rmorshea:
@Archmonger, at the moment, this change proposes that all messages between the server and client contain a `type` key. For example:

```python
{
    "type": "layout-event",
    ...
}
```

I'm open to alternatives. For example, adding `header` key to each message which is an object containing things like the message type and version. This has the advantage of reserving a namespace under which other useful information could be added in the future. However, I am worried about keeping message sizes as small as possible and adding a bunch of redundant boilerplate to message headers could introduce a base level of unavoidable latency. My thinking with the single `type` key is that in the future, if more information is needed, any of the following could be done:

- Version information could be communicated either by encoding it into the type string (e.g. `layout-event@v1.0`)
- Extra metadata could be defining new messages types with that added information.
- Arbitrary message metadata could be communicated by defining handshake message types that the client/server open their communication with and that they use to declare what message types and versions they use. (this seems preferable)

Archmonger:
I support this configuration, except for encoding the version into the type string (`layout-event@v1.0`).

I would suggestion this format

```python
{
    "type": "layout-event",
    "version": 1,  # `version` should be optional. If undefined, assume latest version of the API.
    ...
}

```

Issue #760, opened 2022-06-13
Do Not Unmount Old View On Reconnect


rmorshea:
### Current Situation

Presently, upon reconnecting, the client [unmounts the old view and mounts a new one](https://github.com/idom-team/idom/blob/d1a69f676be43da15c9d2c7c419f55e4acb10ef5/src/client/packages/idom-client-react/src/mount.js#L42). This results in a brief, but jarring flash as the new view loads.

### Proposed Actions

Instead of unmounting the old view we can simply retain it. The first message from the server should be the new VDOM which should be set at the root. This will cause React to render the full tree of components, but it will prevent the view from flashing from the user's perspective.

Archmonger:
Could not unmounting the view cause a memory leak?

rmorshea:
I don't think so. As a note for this issue, it looks like `ReactDOM.render` is deprecated in favor of `createRoot` which returns an object with `render` and `unmount` methods. This seems to rather explicitly allow for `render` to be called multiple times as we'd want to do here.

Archmonger:
This issue was resolved in #886 and #901.

There is still the separate issue of the initial `mount` deleting the inner contents of the div, which can complicate #578.

Closing this issue.

PR #886, opened 2023-01-24
no need to unmount on reconnect

rmorshea:
closes: #760 

I was unable to get preact working with the new React `createRoot` API. We can handle that in a follow-up issue. This narrowly focuses on removing an unnecessary unmount call on reconnect.

## Checklist

### DIFF
--- a/src/client/packages/idom-client-react/src/mount.js
+++ b/src/client/packages/idom-client-react/src/mount.js
@@ -38,9 +38,6 @@ function mountLayoutWithReconnectingWebSocket(
   socket.onopen = (event) => {
     console.info(`IDOM WebSocket connected.`);
 
-    if (mountState.everMounted) {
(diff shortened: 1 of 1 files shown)


Archmonger:
I think a consequence of this change is that now stale event handlers can stay within the client.

This exception occurs if I restart the webserver while keeping an IDOM tab open:

```
2023-01-28T23:57:45-0800 | INFO | Ignored event - handler '3d35445264fb4d79a821e59f5961fc43' does not exist or its component unmounted
```

rmorshea:
yup, that sounds right

Archmonger:
I don't think this is an intended though, since it silently breaks all event handlers on reconnect.

Can we potentially refresh all handlers within the client on reconnect?

rmorshea:
oh, i was agreeing that it sounds like a bug.

PR #901, opened 2023-02-01
Fix unmount

### LINKED ISSUE #896: Stale event handlers can reside in the client
### Current Situation

A consequence of [this change](https://github.com/idom-team/idom/pull/886) is that now stale event handlers can stay within the client.

To replicate, restart the ASGI webserver while keeping an IDOM tab open. This will silently breaks all event handlers on reconnect.

```
2023-01-28T23:57:45-0800 | INFO | Ignored event - handler '3d35445264fb4d79a821e59f5961fc43' does not exist or its component unmounted
```

### Proposed Actions

Look into the following
- Is there a way of unmounting a React component that doesn't delete the DOM tree?
   - We might want to peek at the `ReactDOM.unmountComponentAtNode` source code and determine if we can do this 
- Can we potentially refresh all handlers within the client on reconnect?
- Do we need to revert this PR?

rmorshea:
closes: #896 

Ignore the white space changes due to latest version of black

## Checklist

- [X] The `changelog.rst` has been updated with any significant changes.
- [X] GitHub Issues which may be closed by this Pull Request have been linked.

### DIFF
--- a/docs/source/about/changelog.rst
+++ b/docs/source/about/changelog.rst
@@ -23,7 +23,13 @@ more info, see the :ref:`Contributor Guide <Creating a Changelog Entry>`.
 Unreleased
 ----------
 
-No changes.
+**Reverted**
+
+- :pull:`901` - reverts :pull:`886` due to :issue:`896`
+
+**Fixed**
+
+- :issue:`896` - Stale event handlers after disconnect/reconnect cycle
 
 
 v1.0.0-a1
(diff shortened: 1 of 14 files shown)


rmorshea:
Will reopen #760 on merge

Archmonger (approved):

docs/source/about/changelog.rst
```diff
@@ -23,7 +23,13 @@ more info, see the :ref:`Contributor Guide <Creating a Changelog Entry>`.
 Unreleased
 ----------
 
-No changes.
+**Reverted**
```
Archmonger, line 26: Just realized we're using a non-standard `Keep-A-Changelog` category here. Not a big deal though.

PR #911, opened 2023-02-03
fix data-* and aria-* attribute name conversion

### LINKED ISSUE #912: `data-*` string mutations are broken
> `data_bs_toggle="..."` is rendered as `data-bs_toggle="..."` (idom==1.0.0a2).

_Originally posted by @numpde in https://github.com/idom-team/idom/discussions/903#discussioncomment-4862494_

rmorshea:
Stems from [this comment](https://github.com/idom-team/idom/discussions/903#discussioncomment-4862494).

## Checklist

### DIFF
--- a/src/client/packages/idom-client-react/src/element-utils.js
+++ b/src/client/packages/idom-client-react/src/element-utils.js
@@ -64,7 +64,7 @@ function normalizeAttribute([key, value]) {
     key.startsWith("aria_") ||
     DASHED_HTML_ATTRS.includes(key)
   ) {
-    normKey = key.replace("_", "-");
+    normKey = key.split("_").join("-");
   } else {
     normKey = snakeToCamel(key);
   }
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -142,6 +142,7 @@ async def test_data_and_aria_attribute_naming(display: DisplayFixture):
             id="title",
             data_some_thing="some data",
             aria_description="some title",
+            **{"data-another-thing": "more data"},
         )
 
     await display.show(SomeComponent)
@@ -149,4 +150,5 @@ async def test_data_and_aria_attribute_naming(display: DisplayFixture):
     title = await display.page.wait_for_selector("#title")
 
     assert await title.get_attribute("data-some-thing") == "some data"
+    assert await title.get_attribute("data-another-thing") == "more data"
     assert await title.get_attribute("aria-description") == "some title"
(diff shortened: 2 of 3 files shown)


Archmonger (dismissed):

src/client/packages/idom-client-react/src/element-utils.js
```diff
@@ -64,7 +64,7 @@ function normalizeAttribute([key, value]) {
     key.startsWith("aria_") ||
     DASHED_HTML_ATTRS.includes(key)
   ) {
-    normKey = key.replace("_", "-");
```
Archmonger, line 67: Why doesn't this work though

tests/test_client.py
```diff
@@ -132,3 +132,21 @@ async def handle_change(event):
...
+
+
+async def test_data_and_aria_attribute_naming(display: DisplayFixture):
+    @idom.component
+    def SomeComponent():
+        return idom.html.h1(
+            "see attributes",
+            id="title",
+            data_some_thing="some data",
+            aria_description="some title",
```
Archmonger, line 144: This could also include a test for `**{"data-test-attr": "some data"}`

src/client/packages/idom-client-react/src/element-utils.js
Archmonger, line 67: An alternative to splitting the string is using `replaceAll`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll

rmorshea: Dang stack overflow lied to me. Will switch to that.

Archmonger: To be fair, replaceAll was standardized fairly recently (mid-2020).

Archmonger: Given `replaceAll` was only fully adopted by all browsers early 2021...

For compatibility purposes, it might make sense to keep using `key.split("_").join("-")`

rmorshea: We can add a comment to this effect.

numpde: Maybe a __ could stand for an actual _ in the attribute name. Also, the regex in snakeToCamel could be more permissive, e.g., uppercase anything that's [allowed](https://www.w3.org/TR/2011/WD-html5-20110525/syntax.html#attributes-0), cf. comment [here](https://stackoverflow.com/a/926136/). Or, normalize the attribute name (i.e., "project" to ascii) and complain if not equal.

Archmonger:
This PR seems moot, since aria attributes are going to be handled via dashes `-` after reverting to the old API.

PR #919, opened 2023-02-10
revert #841

### LINKED ISSUE #920: Change `key` API for `idom.html` and/or `idom.component` elements
See [original discussion.](https://github.com/idom-team/idom/discussions/916#discussioncomment-4927208)

rmorshea:
closes: #920

and reverts changes from #841 as per discussion in #916

but preserve ability to declare attributes with snake_case

## Checklist


src/idom/core/vdom.py
```diff
@@ -133,11 +130,12 @@ def is_vdom(value: Any) -> bool:
...
-    *children: VdomChild | VdomChildren,
-    key: Key | None = None,
-    **attributes: Any,
+    *attributes_and_children: VdomAttributesAndChildren,
+    key: str | int | None = None,
+    event_handlers: EventHandlerMapping | None = None,
+    import_source: ImportSourceDict | None = None,
 ) -> VdomDict:
-    """A helper function for creating VDOM elements.
+    """A helper function for creating VDOM dictionaries.
```
Archmonger: Probably want to keep this docstring change

scripts/fix_vdom_constructor_usage.py
```diff
@@ -1,370 +0,0 @@
-from __future__ import annotations
```
Archmonger, line 1: Might want to keep this file, depending on whether we want to auto migrate `key=...`

src/idom/core/vdom.py
```diff
@@ -251,20 +233,44 @@ def constructor(
...
 
     for k, v in attributes.items():
         handler: EventHandlerType
 
         if callable(v):
             handler = EventHandler(to_event_handler_function(v))
         elif (
-            # isinstance check on protocols is slow - use function attr pre-check as a
-            # quick filter before actually performing slow EventHandlerType type check
+            # isinstance check on protocols is slow, function attr check is a quick filter
```
Archmonger: might want to revert this comment change

src/idom/html.py
```diff
@@ -392,20 +406,29 @@ def script(*children: str, key: Key | None = None, **attributes: Any) -> VdomDic
     function that is called when the script element is removed from the tree, or when
     the script content changes.
     """
+    model: VdomDict = {"tagName": "script"}
+
+    attributes, children = coalesce_attributes_and_children(attributes_and_children)
```
Archmonger: We might want to consider renaming `coalesce` to `split`?

src/idom/utils.py
```diff
@@ -265,6 +266,12 @@ def _generate_vdom_children(
     )
 
 
+def _hypen_to_camel_case(string: str) -> str:
```
Archmonger: We probably don't want this behavior anymore.

It's possible some users want to use custom props such as `my-value=...` that contain hyphens. It's technically not a HTML5 violation. So ideally, we shouldn't touch hyphens.

tests/test_core/test_serve.py
```diff
@@ -11,7 +11,7 @@
 from tests.tooling.common import event_message
 
 
-EVENT_NAME = "on_event"
+EVENT_NAME = "onEvent"
```
Archmonger: Change needs confirmation

.github/workflows/test.yml
```diff
@@ -23,7 +23,7 @@ jobs:
       session-name: test_python_suite
       session-arguments: --maxfail=3 --no-cov
       runs-on-array: '["ubuntu-latest", "macos-latest", "windows-latest"]'
-      python-version-array: '["3.7", "3.8", "3.9", "3.10", "3.11"]'
+      python-version-array: '["3.7", "3.8", "3.9", "3.10"]'
```
Archmonger: Need to re-add 3.11 tests

.pre-commit-config.yaml
```diff
@@ -1,14 +1,14 @@
...
   - repo: https://github.com/pycqa/isort
-    rev: 5.6.3
+    rev: 5.12.0
     hooks:
       - id: isort
         name: isort
   - repo: https://github.com/pre-commit/mirrors-prettier
-    rev: "v2.5.1"
+    rev: v2.7.1
     hooks:
```
Archmonger, line 13: un-revert these bumps

scripts/fix_vdom_constructor_usage.py
rmorshea: I think it'd probably be simpler to just have a normal deprecation warning for that. I can always go back in the history to find this as well.

.pre-commit-config.yaml
rmorshea: These were manually added. Was having some trouble with pre-commit for some reason.

scripts/fix_vdom_constructor_usage.py
rmorshea: I'm actually going to handle moving key to be an attribute. There's a lot of places in the docs I don't want to fix so I may actually write a script to auto-update those things.

docs/examples.py
```diff
@@ -122,7 +122,7 @@ def Wrapper():
     def PrintView():
         text, set_text = idom.hooks.use_state(print_buffer.getvalue())
         print_buffer.set_callback(set_text)
-        return idom.html.pre(text, class_name="printout") if text else idom.html.div()
+        return idom.html.pre({"class": "printout"}, text) if text else idom.html.div()
```
Archmonger, line 125: `"class"` needs a revert to `"class_name"`. Also confirm throughout the docs we aren't missing any of those situations.

docs/source/about/changelog.rst
```diff
@@ -23,7 +23,15 @@ more info, see the :ref:`Contributor Guide <Creating a Changelog Entry>`.
...
-No changes.
+**Changed**
+
+- :pull:`919` - Reverts :pull:`841` as per the conclusion in :discussion:`916`. but
+  preserves the ability to declare attributes with snake_case.
+
+**Deprecated**
+
+- :pull:`919` - Declaration of keys via keywork arguments in standard elements. A script
+  has been added to automatically convert old usages where possible.
```
Archmonger, line 34: Do we want to expand this to components as well, so we can ultimately fix our component type hints?

docs/source/guides/adding-interactivity/components-with-state/_examples/isolated_state/main.py
```diff
@@ -29,14 +29,14 @@ def handle_more_click(event):
...
+        html.button({"onClick": handle_next_click}, "Next"),
         html.h2(name, " by ", artist),
         html.p(f"({bounded_index + 1} or {len(sculpture_data)})"),
-        html.img(src=url, alt=alt, style={"height": "200px"}),
+        html.img({"src": url, "alt": alt, "style": {"height": "200px"}}),
         html.div(
             html.button(
-                f"{('Show' if show_more else 'Hide')} details",
-                on_click=handle_more_click,
+                {"onClick": handle_more_click},
```
Archmonger: Need to confirm throughout the docs we are using the new underscore syntax.

Probably need to update the script to handle this.

src/idom/_console/rewrite_key_declarations.py
```diff
@@ -22,7 +21,7 @@
 
 @click.command()
 @click.argument("paths", nargs=-1, type=click.Path(exists=True))
-def update_html_usages(paths: list[str]) -> None:
+def rewrite_key_declarations(paths: list[str]) -> None:
```
Archmonger: maybe just `rewrite_keys` instead?

If we want this to handle the underscore syntax for props, this might become `v1_syntax_update`

rmorshea:
@Archmonger I think we can go forward with the rename after this.

src/idom/core/hooks.py
```diff
@@ -281,7 +277,7 @@ def __init__(
 
     def render(self) -> VdomDict:
         current_hook().set_context_provider(self)
-        return vdom("", *self.children)
+        return {"tagName": "", "children": self.children}
```
Archmonger, line 280: Is there a reason this was switch to a raw VDOM dict?

src/idom/core/hooks.py
```diff
@@ -144,15 +140,15 @@ def use_effect(
...
+    last_clean_callback: Ref[_EffectCleanFunc | None] = use_ref(None)
 
     def add_effect(function: _EffectApplyFunc) -> None:
         if not asyncio.iscoroutinefunction(function):
             sync_function = cast(_SyncEffectFunc, function)
         else:
             async_function = cast(_AsyncEffectFunc, function)
 
-            def sync_function() -> Optional[_EffectCleanFunc]:
+            def sync_function() -> _EffectCleanFunc | None:
```
Archmonger, line 151: We might want to add `pyupgrade` as a precommit-ci hook to automagically handle these

src/idom/html.py
```diff
@@ -280,7 +280,18 @@
 
 def _(*children: Any, key: Key | None = None) -> VdomDict:
     """An HTML fragment - this element will not appear in the DOM"""
-    return vdom("", *children, key=key)
+    attributes, coalesced_children = separate_attributes_and_children(children)
```
Archmonger: Need to rename `coalesced_children` to `children`. We should do a global search on `coalesce` just to clean up references.

src/idom/html.py
```diff
@@ -392,20 +406,29 @@ def script(*children: str, key: Key | None = None, **attributes: Any) -> VdomDic
     function that is called when the script element is removed from the tree, or when
```
Archmonger, line 414: Unrelated to this PR...
We should add a disclaimer here to warn users not to use raw user inputs (or any potentially untrusted data source) within the script contents to avoid `XSS` attacks.

src/idom/utils.py
```diff
@@ -309,6 +308,9 @@ def _vdom_attr_to_html_str(key: str, value: Any) -> tuple[str, str]:
     return key.lower(), str(value)
 
 
+# Pattern for delimitting camelCase names (e.g. camelCase to camel-case)
+_CAMEL_CASE_SUB_PATTERN = re.compile(r"(?<!^)(?=[A-Z])")
+
 # see list of HTML attributes with dashes in them:
 # https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes#attribute_list
-_DASHED_HTML_ATTRS = {"accept_charset", "http_equiv"}
+_DASHED_HTML_ATTRS = {"acceptCharset", "httpEquiv"}
```
Archmonger: tbh these globals are innocent enough to be public instead of private.

It's possible some external libraries might want to rely on it some day, despite it being undocumented.

tests/test_core/test_component.py
```diff
@@ -35,19 +35,19 @@ def SimpleParamComponent(tag):
 async def test_component_with_var_args():
     @idom.component
     def ComponentWithVarArgsAndKwargs(*args, **kwargs):
-        return idom.html.div(*args, **kwargs)
+        return idom.html.div(kwargs, args)
 
-    assert ComponentWithVarArgsAndKwargs("hello", "world", my_attr=1).render() == {
+    assert ComponentWithVarArgsAndKwargs("hello", "world", myAttr=1).render() == {
```
Archmonger: Is the camelCase intentional here?

src/idom/core/hooks.py
rmorshea: That would be great to have

tests/test_core/test_component.py
rmorshea: Nope, not sure why that was camel in the first place given this is mostly a reversion.

src/idom/core/hooks.py
Archmonger: I have a pretty exhaustive "zero effort auto fixes" precommit-ci config in Conreq's App store branch, if you want to take a peek.

https://github.com/Archmonger/Conreq/blob/d3740f4b7a340491c886dd114b7f2f40c398c4a1/.pre-commit-config.yaml

src/idom/core/hooks.py
rmorshea: Just faster. No particular reason.

docs/source/about/changelog.rst
rmorshea: I'm not quite sure how doing this would fix component type hints. Also components have a completely different interface to standard elements - they're effectively a normal function call with the exception of the reserved `key` parameter.

src/idom/core/hooks.py
rmorshea: Feel free to make a PR for this. I don't have time to check this out at the moment.

src/idom/html.py
rmorshea: Can you create an issue for this?

docs/source/guides/adding-interactivity/components-with-state/_examples/isolated_state/main.py
rmorshea: Yup. I think we can do this in a separate PR. I guess we have one more before renaming.

rmorshea: Granted using camelCase is completely valid.

src/idom/core/hooks.py
Archmonger: Sure, I'll PR after v1.

docs/source/about/changelog.rst
Archmonger: I recall us discovering that our decorator can preserve type hints the original function's `*args, *kwargs` if we aren't trying to add a new kwarg (`key=...`).

rmorshea: That's true, though we'd need a wrapper object around keys in order to distinguish them from other positional arguments. We could implement that as shown in [this gist](https://gist.github.com/5938469562e51a8c4617a8f472c4c684), but we should definitely do that in a separate PR though. Perhaps an opinion from @numpde here would be useful too.

rmorshea: This also opens up other questions. Should we allow or require keys for standard elements to be declared in a similar manner? That is, as an optional first positional argument instead of via the normal attributes dict?

docs/examples.py
rmorshea:
```suggestion
        return idom.html.pre({"class_name": "printout"}, text) if text else idom.html.div()
```

docs/source/about/changelog.rst
Archmonger: > as an optional first positional argument instead of via the normal attributes dict

I think you might have mistyped

EDIT: Oh I think I'm getting what you meant. As in, should the `Key` class be usable with `idom.html` as well.

I guess the idea is, if we're doing `Key(...)` just for components, then at that point the props dict on `idom.html` might feel weird 🙃

But they both have caveats... so I would say let's continue with the current `dict` for `idom.html` and `kwarg` for `@component` for now. If we think of a better way of doing it down the line, that'll be a different major version release.

docs/source/reference/_examples/simple_dashboard.py
```diff
@@ -83,10 +83,10 @@ def update_value(value):
         set_value_callback(value)
 
     return idom.html.fieldset(
-        idom.html.legend(label, style={"font_size": "medium"}),
+        {"class": "number-input-container"},
```
Archmonger, line 86: `class` -> `class_name`

rmorshea:
```suggestion
        {"class_name": "number-input-container"},
```

rmorshea:
I'd like to get this out, so I'm going to merge and do a pre-release

Archmonger:
I'll do a post-merge review tomorrow. If there's any issues we can handle it in a separate release.

src/idom/_console/rewrite_camel_case_props.py
```diff
@@ -0,0 +1,88 @@
...
+    rewrite_changed_nodes,
+)
+
+
+CAMEL_CASE_SUB_PATTERN = re.compile(r"(?<!^)(?=[A-Z])")
+
+
+@click.command()
+@click.argument("paths", nargs=-1, type=click.Path(exists=True))
+def rewrite_camel_case_props(paths: list[str]) -> None:
```
Archmonger, line 23: This has issues. Currently it does not preserve dashes `-`, so `data-success` gets turned into `data_success`.

Also it adds an empty `VdomAttributes` dict no reason, For example `html.hr()` becomes `html.hr({})`

src/idom/_console/rewrite_keys.py
```diff
@@ -0,0 +1,111 @@
...
+from idom._console.ast_utils import (
+    ChangedNode,
+    find_element_constructor_usages,
+    rewrite_changed_nodes,
+)
+
+
+@click.command()
+@click.argument("paths", nargs=-1, type=click.Path(exists=True))
+def rewrite_keys(paths: list[str]) -> None:
```
Archmonger, line 19: We probably want to combine both of these into a more general `v1_syntax_update` command. I think that name makes sense, since we're likely to delete these commands once we reach v2.

src/idom/_console/rewrite_camel_case_props.py
rmorshea: `data_success` will be handled properly. I think I did that explicitly to have uniform styling. I can change that though. The latter case is definitely an issue.

Archmonger: Yeah, I think `data-` and `aria-` shouldn't be auto-converted. It's fine if IDOM's client can automatically interpret underscored `data_` attributes, but the suggested styling should remain as a dash since our API allows it.

docs/examples.py
Archmonger: This appears to still be unresolved?

docs/source/_exts/custom_autosectionlabel.py
```diff
@@ -30,6 +30,7 @@ def get_node_depth(node: Node) -> int:
 
 def register_sections_as_label(app: Sphinx, document: Node) -> None:
     docname = app.env.docname
+    print(docname)
```
Archmonger, line 33: Unneeded print?

docs/source/_exts/custom_autosectionlabel.py
```diff
@@ -66,7 +67,7 @@ def register_sections_as_label(app: Sphinx, document: Node) -> None:
         domain.labels[name] = docname, labelid, sectname
 
 
-def setup(app: Sphinx) -> dict[str, Any]:
+def setup(app: Sphinx) -> Dict[str, Any]:
```
Archmonger, line 70: No need to use `typing.Dict` here.

docs/source/guides/adding-interactivity/dangers-of-mutability/_examples/moving_dot_broken.py
```diff
@@ -14,25 +14,29 @@ def handle_pointer_move(event):
...
-                "border_radius": "50%",
-                "width": "20px",
-                "height": "20px",
-                "left": "-10px",
-                "top": "-10px",
-                "transform": f"translate({position['x']}px, {position['y']}px)",
+            {
+                "style": {
+                    "position": "absolute",
+                    "backgroundColor": "red",
```
Archmonger, line 30: Should we convert styles to `snake_case`?

docs/source/guides/adding-interactivity/multiple-state-updates/_examples/set_color_3_times.py
```diff
@@ -15,9 +15,11 @@ def handle_reset(event):
 
     return html.div(
         html.button(
-            "Set Color", on_click=handle_click, style={"background_color": color}
+            {"on_click": handle_click, "style": {"backgroundColor": color}}, "Set Color"
```
Archmonger, line 18: Tons of `backgroundColor` references in the docs. We might want them to be `background_color`.

tests/test_core/test_serve.py
```diff
@@ -115,8 +115,8 @@ async def handle_event():
             second_event_did_execute.set()
 
         return idom.html.div(
-            idom.html.button(on_click=block_forever),
-            idom.html.button(on_click=handle_event),
+            idom.html.button({"onClick": block_forever}),
```
Archmonger, line 118: `onClick` -> `on_click`
There's a few instances of unconverted values in the tests. This might indicate the conversion script isn't working in some edge cases.

docs/source/guides/adding-interactivity/state-as-a-snapshot/_examples/print_chat_message.py
```diff
@@ -16,24 +16,28 @@ async def handle_submit(event):
         print(f"Sent '{message}' to {recipient}")
 
     return html.form(
+        {"on_submit": handle_submit, "style": {"display": "inline-grid"}},
         html.label(
+            {},
```
Archmonger, line 21: There's a handful of these empty dicts in this PR.

I recommend doing a search through this diff page for `{}`

Archmonger:
Been testing this out in Conreq. I have a couple of comments

1) Using `custom_vdom_constructor` on `_` and `script` elements causes pylint errors 
   - We should probably handle this through overloads and helper functions instead.
   - image

2) On some of my `idom.html` elements, I was constructing the vdom dict using function calls (ex. `generate_props(val_1, val_2)`). The `rewrite-keys` script then would add another vdom dict above it, ultimately breaking things.
   - Should we update the script to run a type-check on the first arg of an `idom.html` call (and then combine them with the pipe `|` operator if needed)?
   - The following error probably needs more verbosity in order to better catch things like this:

```
  File "C:\Users\Markg\Documents\Repositories\Conreq\.venv\lib\site-packages\idom\core\layout.py", line 202, in _render_model
    new_state.model.current = {"tagName": raw_model["tagName"]}
KeyError: 'tagName'
```

src/idom/_console/rewrite_keys.py
Archmonger: not needed so I'm closing this one

docs/source/guides/adding-interactivity/multiple-state-updates/_examples/set_color_3_times.py
rmorshea: I think I'm just going to go in and manually change these at this point unless you need this too. I think we can plan on removing these rewrite scripts after the name change anyway.

rmorshea:
Unfortunately I don't think there's a good great way handle that case you're experiencing. That first posarg could also be a child rather than a props dict. I think we can just produce a more helpful error.

Archmonger:
Yeah no problem. To be honest it probably doesn't need the extra robustness, since as far as I can tell I'm the only one "stress testing" IDOM.

The error improvement would be a good addition though

PR #936, opened 2023-02-23
Rewrite style props

rmorshea:
## Checklist

### DIFF
--- a/docs/source/reference/_examples/simple_dashboard.py
+++ b/docs/source/reference/_examples/simple_dashboard.py
@@ -83,7 +83,7 @@ def NumberInput(label, value, set_value_callback, domain):
         set_value_callback(value)
 
     return idom.html.fieldset(
-        {"class": "number-input-container"},
+        {"class_": "number-input-container"},
         idom.html.legend({"style": {"font-size": "medium"}}, label),
         Input(update_value, "number", value, attributes=attrs, cast=float),
         Input(update_value, "range", value, attributes=attrs, cast=float),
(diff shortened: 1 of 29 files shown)


docs/source/reference/_examples/simple_dashboard.py
```diff
@@ -83,7 +83,7 @@ def update_value(value):
         set_value_callback(value)
 
     return idom.html.fieldset(
-        {"class": "number-input-container"},
+        {"class_": "number-input-container"},
```
Archmonger: Is this correct?

rmorshea: Nope, good catch.

Archmonger (approved):

Archmonger:
Do we want to tackle the [`custom_vdom_constructor` pylint issues](https://github.com/idom-team/idom/pull/919#issuecomment-1441081230)?

rmorshea:
So `custom_vdom_constructor` actually does implement an overload. My suspicion is that `pylint` assumes decorators don't modify function interfaces - hence the error messages. IMO, if you're using MyPy you don't need PyLint. If you want to continue using it, I'd recommend just turning off this check.

Archmonger:
Pylint catches a lot of things that I've seen mypy miss. Typically that's a fairly useful error to retain.

It seems like a bug within Pylint, and the generally accepted solution is to directly call the decorator, rather than using the `@` decorator operation.

https://stackoverflow.com/questions/40135129/pylint-getting-it-to-understand-decorators

PR #943, opened 2023-02-27
Fix `idom.run` uvicorn self.servers exception

Archmonger:
This PR makes `idom.run` utilize the same code as `uvicorn.run` in order to prevent unexpected exceptions.

See https://github.com/encode/uvicorn/pull/1845 for details

## Checklist


src/reactpy/backend/_common.py
```diff
@@ -31,32 +30,18 @@ async def serve_development_asgi(
...
-    try:
-        await asyncio.gather(*coros)
-    finally:
-        await asyncio.wait_for(server.shutdown(), timeout=3)
-
-
-async def _check_if_started(server: UvicornServer, started: asyncio.Event) -> None:
-    while not server.started:
-        await asyncio.sleep(0.2)
-    started.set()
```
rmorshea, line 59: You need a way to trigger the `started` event so the tests know when the server is available.

Archmonger: Why are the tests utilizing the `idom.run` API at all?

rmorshea: This isn't the run API. It's a [`serve_development_app`](https://github.com/reactive-python/reactpy/blob/main/src/reactpy/backend/types.py#L33-L40) function that all the backends expose. `run()` does happen to use this same function, but we'd need something like this in order to make writing a common set of tests for each backend easier.

Archmonger:
The [`fix nox`](https://github.com/reactive-python/reactpy/commit/79dcb7bd2c8a300aeeb37e662b4f4ef12ad50da1) commit doesn't seem to have fixed nox.

On a tangent, we should probably set `main` as a protected branch to prevent direct commits.

rmorshea:
Yeah, my bad. I deleted the old ReactPy releases not realizing that once I deleted the last one the name would become available again so was stressing trying to get a release out ASAP.

rmorshea:
Also, `main` is protected. It's just that, as the owner of the repo I'm able to bypass those restrictions.

Archmonger:
@rmorshea Do you agree with removing `VictoryBar` tests due to our `module_from_template` deprecation? That test is super flaky anyways.

src/reactpy/backend/_common.py
```diff
@@ -31,29 +30,32 @@ async def serve_development_asgi(
...
 
+    # If a started event is provided, then use it signal based on `server.started`
     if started:
         coros.append(_check_if_started(server, started))
 
     try:
         await asyncio.gather(*coros)
     finally:
+        if not hasattr(server, "servers"):  # pragma: no cover
+            server.servers = []
```
rmorshea, line 57: What's going on with this line here?

rmorshea:
If we remove the tests we should probably just remove the feature entirely.

src/reactpy/backend/_common.py
Archmonger: Because we aren't using the uvicorn's `run()` API, we can't guarantee uvicorn's order of operations. So we need to make sure `shutdown()` always has an initialized list of `self.servers` to use.

This is mostly due to Uvicorn lacking some variable safety.

See [this issue](https://github.com/encode/uvicorn/pull/1845) for more details.

Archmonger:
I don't feel like this PR was appropriate to remove `module_from_template`, so I backed those changes out.

rmorshea (approved):

PR #947, opened 2023-03-02
More readme link fixes

Archmonger:
- Remove awkward href for the logo
- Fix github actions href

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,11 @@
 <p align="center" id="reactpy">
-    <a href="https://reactpy.herokuapp.com">
-        <img src="https://raw.githubusercontent.com/reactive-python/reactpy/main/branding/svg/reactpy-logo-landscape.svg" alt="ReactPy Logo" style="min-width: 300px; width: 50%" />
-    </a>
+    <img src="https://raw.githubusercontent.com/reactive-python/reactpy/main/branding/svg/reactpy-logo-landscape.svg" alt="ReactPy Logo" style="min-width: 300px; width: 50%" />
 </p>
 <p align="center">
     <em>Reactive user interfaces with pure Python</em>
 </p>
     <p align="center">
-    <a href="https://github.com/encode/starlette/actions">
+    <a href="https://github.com/reactive-python/reactpy/actions">
         <img src="https://github.com/reactive-python/reactpy/workflows/test/badge.svg" alt="Build Status">
     </a>
     <a href="https://pypi.org/project/reactpy/">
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,13 +1,11 @@
 <p align="center" id="reactpy">
-    <a href="https://reactpy.herokuapp.com">
```
rmorshea, line 2: I think clicking the logo ought to take you to the documentation, no?

rmorshea (dismissed):

README.md
Archmonger: It isn't exactly expected that a logo in a GitHub readme is clickable so it would get infrequent use.

Also, would make it a bit awkward for the other repos. For example, directly linking to the ReactPy core docs from the django-reactpy repo would cause confusion. But also changing the link to the django-reactpy docs within that repo also feels weird.

rmorshea: I guess, but images in the readme are clickable by default. It feels weird to see that the logo is clickable and then be taken to the image source. For example, Starlette's logo on their README (which is where I copied the layout from) links the logo to their docs.

I think the issue with linking via the logo in the Django repo is that the logo just says "ReactPy" instead of "ReactPy for Django" or something like that. If it were the latter, it would make sense for it to link to the Django specific docs. With it just being "ReactPy", I agree, that for the Django repo, it's unclear where it should link, if at all.

Archmonger: We can disable the github image link behavior by setting `<a href="#">`.

To be honest, I'm undecided on which option to go with though. But it is a lot more convenient if we just stick with a single, unclickable logo across all repos.

Archmonger: On this note, should we remove the hyperlink on the word `ReactPy` in the description if it might make the `juptyer`/`django`/`plotly` experience feel weird?

rmorshea: I think it needs to be there. The only other link in that paragraph is to ReactJS.

rmorshea: And I'm good removing the link on the image via `#`

Archmonger: Maybe we change that reactpy link to a pypi or GitHub URL instead of docs?

rmorshea: What's the issue with linking to docs?

Archmonger: Nothing particularly wrong with it, but could feel weird in comparison to the ReactJS link?

Without any prior knowledge I would look at that hyperlinked text and assume it goes to a project homepage (such as reactjs.org) which isn't exactly docs.

After thinking about it now, I think that's probably more of an issue with the docs structure not having a real homepage. Maybe the solution is I should create a stylish homepage with the new docs.

rmorshea: I think that's a good observation.

rmorshea (approved):

PR #957, opened 2023-03-17
async effects get prior task and interupt event

### LINKED ISSUE #956: Better Async Effect Cleanup
### Current Situation

```python
@use_effect
    while True:
        await something()
        await something_else()
```

```python
@use_effect
    try:
        while True:
            await something()
            await something_else()
    finally:
        ...  # cleanup logic here
```

### Proposed Actions

```python
@use_effect
    interupt = Event()
    return cleanup.set

    while not interupt.is_set():
        await something()
        await something_else()
    ...  # cleanup logic here
```

```python
@use_effect
    while not interupt.is_set():
        await something()
        await something_else()
    ...  # cleanup logic here
```

rmorshea:
async effects without arguments are now deprecated

## Checklist

### DIFF
--- a/docs/source/about/changelog.rst
+++ b/docs/source/about/changelog.rst
@@ -23,7 +23,7 @@ more info, see the :ref:`Contributor Guide <Creating a Changelog Entry>`.
 Unreleased
 ----------
 
-No changes.
+
 
 
 v1.0.0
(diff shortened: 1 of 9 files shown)


Archmonger:
A lot of misspellings of `interrupt`

rmorshea:
I'm leaning towards deprecating async effects at the moment. I don't think it's possible to defined async effects with a single function. Rather it would probably need to be more like:

```python
effect = use_async_effect()

@effect.setup
async def my_effect_setup(): ...

@effect.teardown
async def my_effect_teardown(task): ...
```

This seems like a different enough interface that we might want to supply it in a separate package (at least initially).

PR #974, opened 2023-05-11
Use span instead of div

rmorshea:
Divs effect how the page is displayed.


rmorshea:
I think this is breaking because of: https://github.com/esm-dev/esm.sh/issues/631

PR #1002, opened 2023-05-29
Poetry+Hatch Monorepo

### LINKED ISSUE #964: Use `poetry` to pin docs dependencies
### Current Situation

Dependencies in the container running the docs are not pinned. This can cause issues with stability.

### Proposed Actions

Use [`poetry`](https://python-poetry.org/) to create a reproducible docs environment in `docs/pyproject.toml`.

rmorshea:
This will set up the project so it can host multiple Python packages. This is in preparation for a future `reactpy-build` package.

Also will introduce Hatch and Poetry for managing Python projects

Closes: #964
Closes: #1002


rmorshea:
@Archmonger I need help debugging [`hatch-build-scripts`](https://github.com/rmorshea/hatch-build-scripts) on windows. I can't seem to figure out what's going wrong just from CI in this PR or [this one in](https://github.com/rmorshea/hatch-build-scripts/pull/2).

Archmonger:
I hopefully can look at this tomorrow. Motherboard just died, my new board arrived today.

Will be installing it tomorrow 🤞

Archmonger:
I just looked at the exception stack in that other PR.

This is actually a very common gotcha. Windows uses backslashes for directory paths, while Linux uses forward slashes.

To fix this, you'll need to remove any hard coded paths in the tests and use the OS-aware `pathlib.Path` instead. If you need string representations then call `str` on the `Path` instance.

rmorshea:
Clearly I'm doing something wrong there, but I don't see where in hatch-build-scripts since I'm using `pathlib`. I kind of just want to see where the files end up when you try and install on windows.

rmorshea:
Ok, turns out it wasn't an issue with `hatch-build-scripts` (though the fact it wasn't raising is a problem). Somehow, by setting `registry-url` in the `setup-node` GH action it expects there to be a `NODE_AUTH_TOKEN` env var available. The env var is set when running `hatch`, but for some reason `hatch` seems to scrub all the environment variables when running build hooks (at least that's what it seems like).

ofek:
> but for some reason hatch seems to scrub all the environment variables when running build hooks (at least that's what it seems like)

It does not actually

rmorshea:
This could very well be some interaction between hatch / invoke / GH actions. Now that I'm thinking about it, if `invoke` doesn't pass on environment variables, that may actually be the culprit.

PR #1008, opened 2023-06-06
report better error for attrs that cannot be serialized

### LINKED ISSUE #437: Explain that component attributes become uninterpreted JSON
This addresses the problems brought up in #347 at their root cause.

### LINKED ISSUE #930: Better Traceback for JSON Serialization Failures
### Discussed in https://github.com/idom-team/idom/discussions/927

<sup>Originally posted by **numpde** February 13, 2023</sup>
The following grim traceback was caused by writing `button(disabled=mutation)` instead of `...=mutation.loading`, and it's representative of a fairly common set. It would be great to have a pointer to near the origin of the error.
```pytb
Exception inside application: Object of type Mutation is not JSON serializable
...
Traceback (most recent call last):
  File "REPO/venv-3.11/lib/python3.11/site-packages/django_idom/websocket/consumer.py", line 48, in disconnect
    await self._idom_dispatcher_future
  File "REPO/venv-3.11/lib/python3.11/site-packages/django_idom/websocket/consumer.py", line 124, in _run_dispatch_loop
    await serve_layout(
  File "REPO/venv-3.11/lib/python3.11/site-packages/idom/core/serve.py", line 41, in serve_layout
    async with create_task_group() as task_group:
  File "REPO/venv-3.11/lib/python3.11/site-packages/anyio/_backends/_asyncio.py", line 662, in __aexit__
    raise exceptions[0]
  File "REPO/venv-3.11/lib/python3.11/site-packages/idom/core/serve.py", line 52, in _single_outgoing_loop
    await send(await layout.render())
  File "REPO/venv-3.11/lib/python3.11/site-packages/channels/generic/websocket.py", line 271, in send_json
    await super().send(text_data=await self.encode_json(content), close=close)
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "REPO/venv-3.11/lib/python3.11/site-packages/channels/generic/websocket.py", line 279, in encode_json
    return json.dumps(content)
           ^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/json/encoder.py", line 200, in encode
    chunks = self.iterencode(o, _one_shot=True)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/json/encoder.py", line 258, in iterencode
    return _iterencode(o, 0)
           ^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/json/encoder.py", line 180, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type Mutation is not JSON serializable
```
</div>

rmorshea:
## Issues

closes: #930
closes: #437

## Summary

In `reactpy.core.vdom.vdom` check if attributes are JSON serializable when in debug mode.

Also misc changes to config options to make this easier in particular ability to specify options as parents.

## Checklist

### DIFF
--- a/src/py/reactpy/reactpy/_option.py
+++ b/src/py/reactpy/reactpy/_option.py
@@ -141,5 +141,14 @@ class DeprecatedOption(Option[_O]):
 
     @Option.current.getter  # type: ignore
     def current(self) -> _O:
-        warn(self._deprecation_message, DeprecationWarning)
+        try:
+            # we access the current value during init to debug log it
+            # no need to warn unless it's actually used. since this attr
+            # is only set after super().__init__ is called, we can check
+            # for it to determine if it's being accessed by a user.
+            msg = self._deprecation_message
+        except AttributeError:
+            pass
+        else:
+            warn(msg, DeprecationWarning)
         return super().current
--- a/src/py/reactpy/reactpy/_option.py
+++ b/src/py/reactpy/reactpy/_option.py
@@ -11,12 +11,6 @@ logger = getLogger(__name__)
 UNDEFINED = cast(Any, object())
 
 
-def deprecate(option: Option[_O], message: str) -> Option[_O]:
-    option.__class__ = _DeprecatedOption
-    option._deprecation_message = message
-    return option
-
-
 class Option(Generic[_O]):
     """An option that can be set using an environment variable of the same name"""
 
@@ -138,7 +132,13 @@ class Option(Generic[_O]):
         return f"Option({self._name}={self.current!r})"
 
 
-class _DeprecatedOption(Option[_O]):  # nocov
+class DeprecatedOption(Option[_O]):
+    """An option that will warn when it is accessed"""
+
+    def __init__(self, *args: Any, message: str, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+        self._deprecation_message = message
+
     @Option.current.getter  # type: ignore
     def current(self) -> _O:
         warn(self._deprecation_message, DeprecationWarning)
--- a/src/py/reactpy/reactpy/_option.py
+++ b/src/py/reactpy/reactpy/_option.py
@@ -8,6 +8,13 @@ from reactpy._warnings import warn
 
 _O = TypeVar("_O")
 logger = getLogger(__name__)
+UNDEFINED = cast(Any, object())
+
+
+def deprecate(option: Option[_O], message: str) -> Option[_O]:
+    option.__class__ = _DeprecatedOption
+    option._deprecation_message = message
+    return option
 
 
 class Option(Generic[_O]):
@@ -16,8 +23,9 @@ class Option(Generic[_O]):
     def __init__(
         self,
         name: str,
-        default: _O | Option[_O],
+        default: _O = UNDEFINED,
         mutable: bool = True,
+        parent: Option[_O] | None = None,
         validator: Callable[[Any], _O] = lambda x: cast(_O, x),
     ) -> None:
         self._name = name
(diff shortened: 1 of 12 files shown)


Archmonger:
We might also want to validate event handlers too. Need to ensure they're either coroutines or callable.

rmorshea:
> We might also want to validate event handlers too.

Event handlers are already extracted from attributes based on whether they are callable or not.

src/py/reactpy/reactpy/_option.py
```diff
@@ -16,8 +23,9 @@ class Option(Generic[_O]):
     def __init__(
         self,
         name: str,
-        default: _O | Option[_O],
```
glaciermountain, line 19: test - delete

src/py/reactpy/reactpy/_option.py
```diff
@@ -8,6 +8,13 @@
 
 _O = TypeVar("_O")
 logger = getLogger(__name__)
+UNDEFINED = cast(Any, object())
+
+
+def deprecate(option: Option[_O], message: str) -> Option[_O]:
+    option.__class__ = _DeprecatedOption
```
glaciermountain: Problem: Directly assigning __class__ is not recommended. Use inheritance instead.

src/py/reactpy/reactpy/_option.py
```diff
@@ -16,8 +23,9 @@ class Option(Generic[_O]):
     def __init__(
         self,
         name: str,
-        default: _O | Option[_O],
+        default: _O = UNDEFINED,
         mutable: bool = True,
+        parent: Option[_O] | None = None,
         validator: Callable[[Any], _O] = lambda x: cast(_O, x),
```
glaciermountain, line 23: The default value for the validator argument is not appropriate. It should be a callable that returns the validated value.

src/py/reactpy/reactpy/_option.py
```diff
@@ -8,6 +8,13 @@
 
 _O = TypeVar("_O")
 logger = getLogger(__name__)
+UNDEFINED = cast(Any, object())
+
+
+def deprecate(option: Option[_O], message: str) -> Option[_O]:
+    option.__class__ = _DeprecatedOption
+    option._deprecation_message = message
+    return option
```
glaciermountain: 1. Modifying the class of object during runtime may cause uncertain and unknown behaviors. It might be good to consider introducing deprecation mechanism as opposed to modifying the class. 
2. Modifying attributes of an object is not the best practise. It would be better to provide appropriate methods to handle those changes.
3. The type hinting for option parameter is incorrect. It should be 'Option[_0] and not just 'Option'. This applies to the return type hint also.

src/py/reactpy/reactpy/_option.py
rmorshea: I'm not too concerned about this. If you'd like, we could make this parameter required in a follow-up PR. Would be happy to accept that contribution.

Archmonger:
> > We might also want to validate event handlers too.
> 
> Event handlers are already extracted from attributes based on whether they are callable or not.

I currently don't see any warnings if I put in a string as an event handler.

```python
from reactpy import component, html, run


@component
def test():
    return html.div({"on_click": "do nothing"}, "Hello World")


run(test)
```

Archmonger:
Also I don't see how this PR resolves #437

rmorshea:
> I currently don't see any warnings if I put in a string as an event handler.

This is strictly about ensuring that attributes are JSON serializable. Know what values are allowed for each attribute on each HTML element would need to be done in a follow-up PR.

rmorshea:
> Also I don't see how this PR resolves https://github.com/reactive-python/reactpy/issues/437

In debug mode, users will get an error if they attempt to pass non-JSON-serializable props to a JS components. They'd get that error without debug mode too, but in debug mode you get a more helpful one.

PR #1024, opened 2023-06-08
Added pre-commit hook 

### LINKED ISSUE #925: Add Back Pre-Commit Hooks
### Current Situation

After repo was re-organized pre-commit was removed just to get that in quicker. We should add back pre-commit hooks.

### Proposed Actions

Add pre-commit hooks that match linting that happens in CI:

- black
- Ruff
- mypy

Smit-Parmar:
## Issues

Resolve #925

## Summary

Added back Pre-Commit Hook 

## Checklist


Evoney (approved):

Evoney (approved):

rmorshea:
A couple areas for improvement:

- It might make sense to have `hatch` actually run pre-commit rather than the other way around. Dong it this way, where pre-commit tells `hatch` to execute the set of linting tools might introduce a fair bit of latency to the typical `git commit` workflow that could be avoided. I could be wrong about that though.
- We'd also like to run the steps that are presently in `lint-js` with pre-commit as well.
- We want pre-commit to fix what it can. Presently that involves passing a `--fix` flag to `hatch run lint-*`

.pre-commit-config.yaml
```diff
@@ -0,0 +1,8 @@
+repos:
+  - repo: local
+    hooks:
```
glaciermountain, line 3: testing

Smit-Parmar:
@rmorshea Opinion on respective point
1.  If pre-commit is running hatch command then it will be automated process whenever someone commits . Latency will be negligible I guess.   
2. Sure we can add lint-js
3. We can add --fix as well

Planning to execute pre-commit in following order
- Fix Python lint.
- Check Python Lint
- Fix JS lint.
- Check JS lint. 
Let me know your thoughts on this.

rmorshea:
We should make it so you don't have to run "fix" and then "check". The "fix" command should return a non-zero exit code if it needs to make corrections. For example, Ruff has a flag for this (`--exit-non-zero-on-fix`). I believe `black` does this by default. Not entirely sure if the JS tools are like Ruff (need a flag) or Black (non-zero on fix by default).

Smit-Parmar:
@rmorshea  Actually I tried using `--exit-non-zero-on-fix` even without using it pre-commit will fail even if code has been auto formatted by `--fix`. 
```
    repo: https://github.com/astral-sh/ruff-pre-commit
    # Ruff version.
    rev: 'v0.0.272'
    hooks:
      - id: ruff
        args: [--fix, --exit-non-zero-on-fix]
```
pre-commit will keep failing unless 4 of the steps completed successfully. What's your suggestion should we make pre-commit like above example (without hatch) or use hatch?

rmorshea:
I'll take a closer look at this shortly.

rmorshea:
I think we can proceed using `hatch run lint-<lang> --fix`. Can you add `pre-commit` as a dependency in the root `pyproject.toml` file?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,29 @@
...
+        entry: hatch run lint-js --fix
+        language: system
+        pass_filenames: false
+  - repo: local
+    hooks:
+      - id: lint-js-check
+        name: Check JS Lint
+        entry: hatch run lint-py
+        language: system
+        pass_filenames: false
```
rmorshea:
```suggestion
        pass_filenames: false

```

rmorshea (approved):

Archmonger:
@rmorshea We still need to enable precommit on the repo

Issue #647, opened 2022-02-06
Improve custom components docs


Archmonger:
### Current Situation

The current docs are a bit sparse on some information, making it hard for new users to develop custom components.

### Proposed Actions

Add more details to the docs.

### Work Items

Add the following information

- Sending event information back to Python-land
- Hello World for JavaScript components
- Intermediate level example (maybe a button?) for JavaScript components

PR #1049, opened 2023-06-14
Possible custom hooks example 647

josephlarrivy:
## Issues

https://github.com/reactive-python/reactpy/issues/647

## Summary

Using reducer and use_reducer, I built a simple example of how I would build a custom hook custom hook in _examples that is able to return multiple functions to another file

## Checklist


rmorshea:
I think the goal of `use_reducer` is to avoid the need to return multiple functions - instead of having a function for each possible operation you might want to perform on your state, you have one function that you pass an "action" which determines how your state will transition.

josephlarrivy:
Alright. I was looking at using this as a simple example to build up to more complicated hooks. I'll have to play around with use_reducer a bit more to get used to using it.

rmorshea:
I'm going to close this for now. Thanks for the submission though!

PR #1050, opened 2023-06-14
Warn and fix missing mime types

Archmonger:
## Issues

N/A

## Summary

Windows registry can [sometimes go corrupt](https://github.com/encode/starlette/issues/829) causing missing mime types. This results in our backends assuming all file types are `text/plain`. Ultimately, this causes us to fail to serve our ReactPy client `index.js`.

This PR aims to only resolve a minimal set of file extensions (only the ones we potentially use within ReactPy core). I considered using a more comprehensive `mime_types` list from [this repo](https://github.com/micnic/mime.json), but it seems a bit overkill for us right now. We may want to use it if we need to serve a variety of static files. 

## Checklist

### DIFF
--- a/docs/source/about/changelog.rst
+++ b/docs/source/about/changelog.rst
@@ -23,7 +23,9 @@ more info, see the :ref:`Contributor Guide <Creating a Changelog Entry>`.
 Unreleased
 ----------
 
-No changes.
+**Fixed**
+
+- :pull:`1050` - Fix issue where JavaScript files can be detected as mime-type ``text/plain`` due to a corrupt Windows registry
 
 
 v1.0.0
--- a/docs/source/about/changelog.rst
+++ b/docs/source/about/changelog.rst
@@ -23,9 +23,9 @@ more info, see the :ref:`Contributor Guide <Creating a Changelog Entry>`.
 Unreleased
 ----------
 
-**Fixed**
+**Changed**
 
-- :pull:`1050` - Fix issue where JavaScript files can be detected as mime-type ``text/plain`` due to a corrupt Windows registry
+- :pull:`1050` - Warn when operating system has missing mime types, which can result in ``reactpy.run`` not working as expected.
 
 
 v1.0.0
--- a/docs/source/about/changelog.rst
+++ b/docs/source/about/changelog.rst
@@ -25,7 +25,7 @@ Unreleased
 
 **Changed**
 
-- :pull:`1050` - Warn when operating system has missing mime types, which can result in ``reactpy.run`` not working as expected.
+- :pull:`1050` - Warn and attempt to fix missing mime types, which can result in ``reactpy.run`` not working as expected.
 
 **Fixed**
 
--- a/src/py/reactpy/reactpy/backend/__init__.py
+++ b/src/py/reactpy/reactpy/backend/__init__.py
@@ -7,12 +7,11 @@ _logger = getLogger(__name__)
 # Example: https://github.com/encode/starlette/issues/829
 if not mimetypes.inited:
     mimetypes.init()
-MIME_TYPES = {
+for extension, mime_type in {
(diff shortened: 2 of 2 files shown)


rmorshea:
I'll review this evening.

src/py/reactpy/reactpy/backend/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+}
+for extension, mime_type in MIME_TYPES.items():
+    if not mimetypes.types_map.get(extension):  # pragma: no cover
+        _logger.warning(
+            "Mime type '%s = %s' is missing. Please research how to "
+            "fix missing mime types on your operating system.",
+            extension,
+            mime_type,
+        )
+        mimetypes.add_type(mime_type, extension)
```
rmorshea: Given that this is a Windows-only issue can we put this in an `if os.name == "nt"` block?

rmorshea (commented):
I wonder if we should be really trying to patch this or just warning about it and linking to resources on the Windows mimetype registry - neither Starlette nor Flask appear to make any similar attempt.

src/py/reactpy/reactpy/backend/__init__.py
Archmonger: It's technically not just a Windows issue, but since the Windows registry is a free-for-all hellhole it's far more likely to happen on Windows.

Archmonger:
I thought the same thing at first.

But the patch is a harmless fix for `reactpy.run`, which is a user's first impression of our framework. If we can improve first impressions with minimal LOCs it's in our best interest to do so.

I didn't want to link to anything specific since the repair method depends on operating system. This issue technically [isn't limited to Windows](https://man7.org/linux/man-pages/man5/mime.types.5.html), but it is far more likely to happen on Windows.

I will update the comment to be more specific about this issue being OS independent.

src/py/reactpy/reactpy/backend/__init__.py
```diff
@@ -0,0 +1,22 @@
+import mimetypes
+from logging import getLogger
+
+_logger = getLogger(__name__)
+
+# Fix for missing mime types due to OS corruption/misconfiguration
```
rmorshea, line 6: Can we link to the starlette/flask issue?

rmorshea:
I just merged a fix for the `test-docs` failure.

src/py/reactpy/reactpy/backend/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+# Example: https://github.com/encode/starlette/issues/829
+if not mimetypes.inited:
+    mimetypes.init()
+MIME_TYPES = {
+    ".js": "application/javascript",
+    ".css": "text/css",
+    ".json": "application/json",
+}
+for extension, mime_type in MIME_TYPES.items():
```
rmorshea: avoids exposing `MIME_TYPES` as something importable and, IMO, doesn't really harm readability.

```suggestion
for extension, mime_type in {
    ".js": "application/javascript",
    ".css": "text/css",
    ".json": "application/json",
}.items():
```

docs/source/about/changelog.rst
```diff
@@ -23,6 +23,10 @@ more info, see the :ref:`Contributor Guide <Creating a Changelog Entry>`.
 Unreleased
 ----------
 
+**Changed**
+
+- :pull:`1050` - Warn when operating system has missing mime types, which can result in ``reactpy.run`` not working as expected.
```
rmorshea: This will also warn and fix the issue if you're running in production.

```suggestion
- :pull:`1050` - Warn, and attempt to fix, when operating system has missing mime types.
```

rmorshea:
After that last comment LGTM

rmorshea (dismissed):

rmorshea (approved):

PR #1051, opened 2023-06-14
`reactpy.run` and `configure(...)` refactoring

Archmonger:
## Issues

Right now `reactpy.run` can fail in fairly unpredictable ways which give little in terms of stack traces. 

## Summary

This refactors the `reactpy.run` call stack in order allow failures that are user-readable.

- Change `reactpy.backends.utils.find_all_implementations()` to first try to import `<backend_name>` before importing `reactpy.backend.<backend_name>`
   - Allows for missing sub-dependencies to not cause `reactpy.run` to silently fail
- Import `uvicorn` directly within `serve_with_uvicorn` in order to defer import.
   - Allows for `ModuleNotFound: Could not import uvicorn` exception to tell the user what went wrong
- Added `CommonOptions.serve_index_route: bool`
   - Allows us to not clutter the route patterns when it's not needed
   - There are real circumstances where a user might want the index route to 404
- Fix bug where in-use ports are being assigned on Windows.
   - Removes `allow_reuse_waiting_ports` parameter on `find_available_port()` 
- Rename `BackendImplementation` to `BackendProtocol`
- Change load order of `SUPPORTED_PACKAGES` so that `FastAPI` has a chance to run before `starlette`
- Rename `SUPPORTED_PACKAGES` to `SUPPORTED_BACKENDS`
- Refactor `reactpy.backend.*` code to be more human readable
- Use f-strings where possible
- Merge `if` statements where possible
- Use `contextlib.supress` where possible
- Remove defunct `requirements.txt` file 

## Checklist

### DIFF
--- a/src/py/reactpy/reactpy/backend/_common.py
+++ b/src/py/reactpy/reactpy/backend/_common.py
@@ -59,10 +59,7 @@ async def serve_with_uvicorn(
         await asyncio.wait_for(server.shutdown(), timeout=3)
 
 
-async def _check_if_started(server, started: asyncio.Event) -> None:
-    import uvicorn
-
-    server: uvicorn.Server = server
+async def _check_if_started(server: Any, started: asyncio.Event) -> None:
     while not server.started:
         await asyncio.sleep(0.2)
     started.set()
--- a/src/py/reactpy/reactpy/backend/_common.py
+++ b/src/py/reactpy/reactpy/backend/_common.py
@@ -24,17 +24,15 @@ STREAM_PATH = PATH_PREFIX / "stream"
 CLIENT_BUILD_DIR = Path(_reactpy_file_path).parent / "_static" / "app" / "dist"
 
 
-
-async def serve_development_asgi(
+async def serve_with_uvicorn(
     app: ASGIApplication | Any,
     host: str,
     port: int,
     started: asyncio.Event | None,
 ) -> None:
     """Run a development server for an ASGI application"""
-
     import uvicorn
-    
+
     server = uvicorn.Server(
         uvicorn.Config(
             app,
@@ -62,7 +60,10 @@ async def serve_development_asgi(
         await asyncio.wait_for(server.shutdown(), timeout=3)
 
 
-async def _check_if_started(server: uvicorn.Server, started: asyncio.Event) -> None:
+async def _check_if_started(server, started: asyncio.Event) -> None:
+    import uvicorn
+
+    server: uvicorn.Server = server
     while not server.started:
         await asyncio.sleep(0.2)
     started.set()
@@ -71,8 +72,7 @@ async def _check_if_started(server: uvicorn.Server, started: asyncio.Event) -> N
 def safe_client_build_dir_path(path: str) -> Path:
     """Prevent path traversal out of :data:`CLIENT_BUILD_DIR`"""
     return traversal_safe_path(
-        CLIENT_BUILD_DIR,
-        *("index.html" if path in ("", "/") else path).split("/"),
+        CLIENT_BUILD_DIR, *("index.html" if path in {"", "/"} else path).split("/")
     )
 
 
--- a/src/py/reactpy/reactpy/backend/_common.py
+++ b/src/py/reactpy/reactpy/backend/_common.py
@@ -109,7 +109,7 @@ def vdom_head_elements_to_html(head: Sequence[VdomDict] | VdomDict | str) -> str
             head = cast(VdomDict, {**head, "tagName": ""})
         return vdom_to_html(head)
     else:
-        return vdom_to_html(html._(head))
+        return vdom_to_html(html._(*head))
 
 
 @dataclass
--- a/src/py/reactpy/reactpy/backend/_common.py
+++ b/src/py/reactpy/reactpy/backend/_common.py
@@ -39,7 +39,6 @@ async def serve_with_uvicorn(
             host=host,
             port=port,
             loop="asyncio",
-            reload=True,
         )
     )
     server.config.setup_event_loop()
--- a/src/py/reactpy/reactpy/backend/_common.py
+++ b/src/py/reactpy/reactpy/backend/_common.py
@@ -23,44 +23,43 @@ STREAM_PATH = PATH_PREFIX / "stream"
 
 CLIENT_BUILD_DIR = Path(_reactpy_file_path).parent / "_static" / "app" / "dist"
 
-try:
+
+
+async def serve_development_asgi(
+    app: ASGIApplication | Any,
+    host: str,
+    port: int,
+    started: asyncio.Event | None,
+) -> None:
+    """Run a development server for an ASGI application"""
+
     import uvicorn
-except ImportError:  # nocov
-    pass
-else:
-
-    async def serve_development_asgi(
-        app: ASGIApplication | Any,
-        host: str,
-        port: int,
-        started: asyncio.Event | None,
-    ) -> None:
-        """Run a development server for an ASGI application"""
-        server = uvicorn.Server(
-            uvicorn.Config(
-                app,
-                host=host,
-                port=port,
-                loop="asyncio",
-                reload=True,
-            )
+    
+    server = uvicorn.Server(
+        uvicorn.Config(
+            app,
+            host=host,
+            port=port,
+            loop="asyncio",
+            reload=True,
         )
-        server.config.setup_event_loop()
-        coros: list[Awaitable[Any]] = [server.serve()]
+    )
+    server.config.setup_event_loop()
+    coros: list[Awaitable[Any]] = [server.serve()]
 
-        # If a started event is provided, then use it signal based on `server.started`
-        if started:
-            coros.append(_check_if_started(server, started))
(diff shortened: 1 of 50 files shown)


Archmonger:
@rmorshea `npm install -g npm@8.5.0` seems to have broken in the github workflow in the last 24 hours. Not sure why, but I tried bumping `actions/setup-node` version. Did not fix it.

rmorshea:
I was planning on doing a 1.0.1 release tomorrow. Should this wait for the next release?

Archmonger:
Yes, technically this release has an API change (`BackendImplementation` to `BackendProtocol`), so we shouldn't put this in a hotfix release.

Archmonger:
I discovered that our `flask` backend doesn't work with `uvicorn` due to `WsgiToAsgi` not allowing [non-HTTP scopes](https://github.com/django/asgiref/blob/78128f185b17b5f23e269bfb052d9d0ffb05c05b/asgiref/wsgi.py#L35). When removing this restriction, this showed that  `flask-sock` relies on via non-standard WSGI extensions which `WsgiToAsgi` does not support.

I attempted to rewrite `WsgiToAsgi` locally but it broke a ton of other stuff.

Ref: https://github.com/miguelgrinberg/flask-sock/issues/58

src/py/reactpy/reactpy/backend/_common.py
```diff
@@ -23,47 +23,43 @@
...
+        # Since we aren't using the uvicorn's `run()` API, we can't guarantee uvicorn's
+        # order of operations. So we need to make sure `shutdown()` always has an initialized
+        # list of `self.servers` to use.
+        if not hasattr(server, "servers"):  # nocov
+            server.servers = []
+        await asyncio.wait_for(server.shutdown(), timeout=3)
 
 
-async def _check_if_started(server: uvicorn.Server, started: asyncio.Event) -> None:
+async def _check_if_started(server: Any, started: asyncio.Event) -> None:
```
rmorshea: You can put the uvicorn import in an `if TYPE_CHECKING` block instead of making this `Any`.

src/py/reactpy/reactpy/backend/default.py
```diff
@@ -5,13 +5,24 @@
...
 
 logger = getLogger(__name__)
+_DEFAULT_IMPLEMENTATION: BackendProtocol[Any] | None = None
 
 
+# BackendProtocol.Options
+class Options:  # nocov
+    """Create configuration options"""
+
+    def __call__(self, *args: Any, **kwds: Any) -> NoReturn:
```
rmorshea:
```suggestion
    def __init__(self, *args: Any, **kwds: Any) -> NoReturn:
```

src/py/reactpy/reactpy/backend/types.py
```diff
@@ -11,11 +11,11 @@
 
 
 @runtime_checkable
-class BackendImplementation(Protocol[_App]):
+class BackendProtocol(Protocol[_App]):
```
rmorshea: While I doubt changing this name will break anything for people we may want to avoid unnecessary renames.

If we do want to rename this, while I haven't been perfect, I've generally named protocols that are object interfaces with a `Type` suffix and those that are callables `Func`. So by that convention we'd name this `BackendType`.

src/py/reactpy/reactpy/backend/utils.py
```diff
@@ -8,62 +8,51 @@
...
     "tornado",
     "flask",
+    "starlette",
 )
 
 
 def run(
     component: RootComponentConstructor,
-    host: str = "127.0.0.1",
+    host: str | None = None,
```
rmorshea: Unclear why this is better.

src/py/reactpy/reactpy/backend/utils.py
```diff
@@ -8,62 +8,51 @@
...
-
     app_cls = type(app)
+
     logger.info(
-        f"Running with {app_cls.__module__}.{app_cls.__name__} at http://{host}:{port}"
+        "ReactPy is running with '%s.%s' at http://%s:%s",
+        app_cls.__module__,
+        app_cls.__name__,
+        host,
+        port,
```
rmorshea, line 46: There's probably a bunch of other places where I used f-string in logs

rmorshea (commented):
Generally just minor things to correct.

src/py/reactpy/reactpy/backend/default.py
rmorshea: The docstring should be updated to reflect that it should not be used too.

src/py/reactpy/reactpy/backend/types.py
Archmonger: Given that `BackendType` is going to have at minimum 4 attributes, and possibly more in the future, should we consider changing our backends to use classes instead?

It feels really disjointed to read right now.

src/py/reactpy/reactpy/backend/utils.py
Archmonger: Yup, I didn't want to clutter this PR too much though. I think covering other files should be a follow up PR.

src/py/reactpy/reactpy/backend/default.py
Archmonger: A docstring that `__init__` should not be used, or am I misunderstanding?

rmorshea: Yeah, the default options should never be constructed.

src/py/reactpy/reactpy/backend/types.py
rmorshea: Given that none of the methods require any shared state (and I'd like to avoid that if possible) I don't think we need these to be attached to classes.

src/py/reactpy/reactpy/backend/tornado.py
```diff
@@ -60,10 +61,12 @@ def configure(
     )
 
 
+# BackendProtocol.create_development_app
```
rmorshea: These comments need to be updated with the new name.

src/py/reactpy/reactpy/backend/types.py
Archmonger: Conceptually they're still tied together though. They could all be static methods in a class perhaps?

When I was looking at creating a backend, it was fairly hideous to stare at our existing files to figure out what's going on.

rmorshea: If things become more complicated we can consider doing that. For now, it seems like it would be similarly complex to make that change without breaking the ability to do `from reactpy.backend.package import configure`.

src/py/reactpy/reactpy/backend/default.py
```diff
@@ -5,13 +5,26 @@
...
 
+# BackendProtocol.Options
+class Options:  # nocov
+    """Create configuration options"""
+
+    def __init__(self, *args: Any, **kwds: Any) -> NoReturn:
+        """This should not be used, since default options should never
+        be constructed. Please override this method."""
+        msg = "Default implementation has no options."
+        raise ValueError(msg)
```
rmorshea: I don't think there's any reason someone should subclass this

```suggestion
# BackendProtocol.Options
class Options:  # nocov
    """Do not instantiate - default implementation has no options"""

    def __init__(self, *args: Any, **kwds: Any) -> NoReturn:
        msg = "Default implementation has no options."
        raise ValueError(msg)
```

src/py/reactpy/reactpy/backend/utils.py
```diff
@@ -8,62 +8,50 @@
...
     for port in range(port_min, port_max):
         with closing(socket.socket()) as sock:
             try:
-                if allow_reuse_waiting_ports:
-                    # As per this answer: https://stackoverflow.com/a/19247688/3159288
-                    # setting can be somewhat unreliable because we allow the use of
-                    # ports that are stuck in TIME_WAIT. However, not setting the option
-                    # means we're overly cautious and almost always use a different addr
-                    # even if it could have actually been used.
-                    sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
```
rmorshea: Not sure why we're deleting this. Without this, if you run the stop and then immediately restart the server, it's very likely to use a different address. This is rather annoying since you can't simply return to your existing browser tab and hit refresh.

Archmonger: Reason is mentioned in the PRs description.

Fixes a bug where in-use ports are being assigned on Windows. Deleting it didn't seem to impact Linux.

rmorshea: Doesn't break on Linux, but every time you restart the server you'll get a different port. Perhaps we can just vary the behavior by platform.

rmorshea: I should double check that, but that is why I recall introducing that.

Archmonger: Okay let me know, if that's behavior on Linux I can just change the default when running win32.

Archmonger:
@rmorshea This seems to be mergeable, but needs approval.

rmorshea (approved):

PR #1056, opened 2023-06-15
Add a template tag

### LINKED ISSUE #653: Embedding ReactPy via Template Tags
### Current Situation

Currently, users are expected to manually construct the component mount HTML needed to embed IDOM into their projects. This is results in poor usability, and even worse, it can very easily cause server/client version mismatches.

### Proposed Actions

We need to increase the convenience embedding ReactPy into existing projects, and document this as the primary way of using IDOM

For any web framework that supports HTML templating, we really should [use a system identical to that used on ReactPy-Django](https://github.com/idom-team/django-idom#my_apptemplatesmy-templatehtml).

The template tag source code will look something like this
```jinja
{% load static %}
<div id="{{ idom_mount_uuid }}" class="{{ class }}"></div>
<script type="module" crossorigin="anonymous">
  import { mountViewToElement } from "{% static 'django_idom/client.js' %}";
  const mountPoint = document.getElementById("{{ idom_mount_uuid }}");
  mountViewToElement(
    mountPoint,
    "{{ idom_websocket_url }}",
    "{{ idom_web_modules_url }}",
    "{{ idom_ws_max_reconnect_timeout }}",
    "{{ idom_component_id }}",
    "{{ idom_component_params }}"
  );
</script>
```

We will probably rely on the [jinja2 simple tags](https://github.com/dldevinc/jinja2-simple-tags) library.

To be framework agnostic, the static URL will probably need to be set via ReactPy's python config.

Archmonger:
*By submitting this pull request you agree that all contributions to this project are made under the MIT license.*

## Issues

- fix #653

## Summary

Adds template tag support for all of our supported backends.

## Checklist

### DIFF
--- a/src/py/reactpy/pyproject.toml
+++ b/src/py/reactpy/pyproject.toml
@@ -41,23 +41,27 @@ all = ["reactpy[starlette,sanic,fastapi,flask,tornado,testing]"]
 starlette = [
   "starlette >=0.13.6",
   "uvicorn[standard] >=0.19.0",
+  "jinja2-simple-tags",
 ]
 sanic = [
   "sanic >=21",
   "sanic-cors",
   "uvicorn[standard] >=0.19.0",
+  "jinja2-simple-tags",
 ]
 fastapi = [
   "fastapi >=0.63.0",
   "uvicorn[standard] >=0.19.0",
+  "jinja2-simple-tags",
 ]
 flask = [
   "flask",
   "markupsafe>=1.1.1,<2.1",
   "flask-cors",
   "flask-sock",
+  "jinja2-simple-tags",
 ]
-tornado = [
+tornado = [ # TODO: Torando does not use Jinja. We will need to write something custom for this.
   "tornado",
 ]
 testing = [
--- a/src/py/reactpy/reactpy/templatetag.py
+++ b/src/py/reactpy/reactpy/templatetag.py
@@ -1,0 +1,39 @@
+from jinja2_simple_tags import StandaloneTag
+from uuid import uuid4
+import urllib.parse
+
+REACTPY_WEBSOCKET_URL: str
+REACTPY_WEB_MODULES_URL: str
+REACTPY_RECONNECT_MAX: str
+REACTPY_CLIENT_URL_PATH: str
+
+class ComponentTag(StandaloneTag):
+    safe_output = True
+    tags = {"component"}
+
+    def render(self, dotted_path: str, *args, **kwargs):
+        uuid = uuid4().hex
+        class_ = kwargs.pop("class", "")
+        kwargs.pop("key", "")  # `key` is effectively useless for the root node
+        url_path = REACTPY_CLIENT_URL_PATH
+        if kwargs.get("args") is not None:
+            raise ValueError("Cannot specify `args` as a keyword argument")
+        if args:
+            kwargs["args"] = args
+        if kwargs:
+            url_path += f"?{urllib.parse.urlencode(kwargs)}"
+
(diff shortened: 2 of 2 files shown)


Archmonger:
@rmorshea this PR contains the template tag itself, but not in a usable form due to
- ~~Needing modification to websocket URL patterns (new pattern: `reactpy-ws-url/<uuid>?<kwargs>`)~~
- Missing [JavaScript mount function](https://github.com/reactive-python/reactpy-django/blob/f2d54e15075109301df0ffc4bd8b267c7f1bfebf/src/js/src/index.js#L13)
- Missing support for Tornado, which [does not use Jinja](https://www.tornadoweb.org/en/stable/template.html)
- Needs proper definitions for the constant vars in `templatetag.py`
- Needs to verify if components have already been registered

I don't know if I'm going to have time to work on this in-between working on the docs, so feel free to commit directly to this branch if you feel like moving it forward.

rmorshea:
This might make more sense as a standalone `reactpy-jinja` package. It also shouldn't require any modification to the current server API. In short, this will require a system similar to Django's in which you registry views and have a single root view that knows how to route requests to the appropriate component.

The current server API accepts WS connections at `_reactp/stream/<path:path>`. As such, all you need to do is render views based on the `path`. That might (very roughly) look something like this:

```python
import json
from urllib.parse import parse_qs, unquote

views: dict[str, Callable[[], Component]] = { ... }

@component
def app():
    loc = use_location()
    cmpt = views[loc.pathname]
    search = json.loads(parse_qs(loc.search[1:])
    args = json.loads(unquote(search.get("args", ["[]"])[0]))
    kwargs = json.loads(unquote(search.get("kwargs", ["{}"])[0]))
    return cmpt(*args, **kwargs)
```

So in the rough pseudo code above, connecting to a WS at:

`_reactpy/stream/my_component?kwargs=%7B%22hello%22%3A%20%22world%22%7D`

would effectively render `views["my_component"](hello="world")`.

One could imagine user's interacting with this library in the following way:

```python
from fastapi import FastAPI
from fastapi.responses import HTMLResponse
from fastapi.templating import Jinja2Templates

from reactpy.backend.fastapi import configure
import reactpy_jinja

@reactpy_jinja.register
def my_component(hello):
    ...

app = FastAPI()

@app.get("/", response_class=HTMLResponse)
async def index(request):
    return templates.TemplateResponse("index.html")

configure(app, reactpy_jinja.root)
```

Where `index.html` uses the `component` tag somewhere:

```html
{% component "my_component" hello="world" %}
```

Archmonger:
You think `reactpy-jinja` could be a good candidate to be contained in the monorepo?

Maybe we should call it `reactpy-template-tag` to allow for future expansion?

The decorator might be finnicky. Unlike ASGI views, users might not have all of their component functions imported into their main file. We should probably require a `template_dirs: list[Glob]` for us to automatically parse components from the available HTML templates. 

Also do we even want to consider anything unique for Tornado, or do we force Tornado users to install jinja? I'm leaning towards the latter.

Archmonger:
I think we really should push for the template tag being built-in to ReactPy. In my opinion, it should be our main way of advertising how to use ReactPy.

rmorshea:
I think this work might be easier once we've consolidated around ASGI/WSGI implementations since it can build upon a more uniform application interface.

Archmonger:
Yep, that's why this issue is marked as blocked for now.

Archmonger:
Template tags are mandatory for `ReactPyMiddleware`, so I will merge the contents of this PR into #1113

Issue #894, opened 2023-01-29
Create a way to evaluate arbitrary javascript


Archmonger:
### Current Situation

Currently, there is no method to transmit the results of a JavaScript function to the backend.

### Proposed Actions

This might look like a `use_script` hook that operates similarly to `html.script`, but can return a result.

For example, 

```python
@component
def example():
    executor = use_script("getComputedStyle(document.querySelector('#element'))")
    return f"CSS Style Attributes: {executor.data}"
```

Since this hook would need to operate asynchronously, the interface would likely be heavily inspired by our Django `use_query` hook.

And perhaps it could be used as an executor within event contexts to be more "reactive":

```python
@component
def example():
    executor = use_script("getComputedStyle(document.querySelector('#element'))", deferred=True)

    async def on_click(event):
        await executor()
        example_value = executor.data 

    return f"CSS Style Attributes: {script.data}"
```

See [this comment](https://github.com/idom-team/idom/issues/857#issuecomment-1359977125) for traceability. 

Note: If possible, we should try pre-populate `this` keyword with something that makes sense, depending on where the JavaScript was executed.

Archmonger:
@rmorshea This issue already exists, may as well continue the conversation here.

PR #1084, opened 2023-07-03
messenger pattern

### LINKED ISSUE #975: Allow server-side components to send messages to client-side components
### Current Situation

Presently, custom client-side components are able to send events back to the server. However, it is not possible to send events from server-side components to client-side ones.

### Proposed Actions

Now that custom JS components have the ability to [register callbacks with the client](https://github.com/reactive-python/reactpy/blob/main/src/client/packages/%40reactpy/client/src/reactpy-vdom.tsx#L22-L25) this should be technologically feasible. With that said, while users can listen in on particular message types, there is no concept of a message "target". We could achieve this by having message types of the form `server-event:<the-target>`, but this seems like a bit of a hack. Perhaps we can allow (require?) a nullable `target` field for this purpose.

Before diving into all those details though, we need to work out exactly what this interface should look like.

I can imagine having an interface similar to:

```python
@component
def example():
    channel = use_channel()

    @use_effect
    async def delayed_message():
        import asyncio
        await asyncio.sleep(5)
        await channel.send({"my": "message"})
        response = await channel.receive()

    return custom_js_component({"channel_id": channel.id})
```

where `custom_js_component` would then subscribe to messages of the type `channel-message` and target `channel.id`.

rmorshea:

## Issues

Currently, while the client and server have the ability to specify different message types, the server has no way of broadcasting messages other than layout events. Doing so is necessary for a number of different use cases...

- https://github.com/reactive-python/reactpy/discussions/1083#discussioncomment-6338943
- https://github.com/reactive-python/reactpy-router/issues/19
- https://github.com/reactive-python/reactpy/issues/894
- Fix #975 

## Solution

Implement a generic `Messenger` class that allows for users to send/receive arbitrary messages to/from the client. Users will gain access to a `Messenger` via the `use_messenger` hook. Usage could look like one of the following...

Using `start_consumer` and `start_producer`:

```python
@component
def demo_producer_consumer():
    msgr = use_messenger()
    use_effect(lambda: msgr.start_consumer("my-message-type", my_message_consumer))
    use_effect(lambda: msgr.start_producer(my_message_producer))

async def my_message_consumer(msg):
    ...

async def my_message_producer():
    while True:
        yield ...
```

Using `send` and `receive`:

```python
@component
def demo_send_receive():
    msgr = use_messenger()

    @use_effect
    async def start_consumer():
        async for msg in msgr.consume("my-message-type"):
            ...
    
    @use_effect
    async def start_producer():
        while True:
            await msgr.send("my-message-type", ...)
```

Ultimately, the `start_consumer` and `start_producer` methods are convenience methods that call `send` and `receive` under the hood.

## Checklist

### DIFF
--- a/src/py/reactpy/reactpy/backend/messenger.py
+++ b/src/py/reactpy/reactpy/backend/messenger.py
@@ -1,0 +1,61 @@
+from __future__ import annotations
+
+from collections.abc import AsyncIterator, Awaitable
+from typing import Callable
+
+from anyio import Event, create_memory_object_stream, create_task_group
+from anyio.abc import ObjectReceiveStream, ObjectSendStream
+
+from reactpy.core.types import Message
+
+_MessageStream = tuple[ObjectSendStream[Message], ObjectReceiveStream[Message]]
+
+
+class Messenger:
+    """A messenger for sending and receiving messages"""
+
+    def __init__(self) -> None:
+        self._task_group = create_task_group()
+        self._streams: dict[str, list[_MessageStream]] = {}
+        self._recv_began: dict[str, Event] = {}
+
+    def start_producer(self, producer: Callable[[], AsyncIterator[Message]]) -> None:
+        """Add a message producer"""
+
+        async def _producer() -> None:
+            async for message in producer():
+                await self.send(message)
+
+        self._task_group.start_soon(_producer)
+
+    def start_consumer(
+        self, message_type: str, consumer: Callable[[Message], Awaitable[None]]
+    ) -> None:
+        """Add a message consumer"""
+
+        async def _consumer() -> None:
+            async for message in self.receive(message_type):
+                self._task_group.start_soon(consumer, message)
+
+        self._task_group.start_soon(_consumer)
+
+    async def send(self, message: Message) -> None:
+        """Send a message to all consumers of the message type"""
+        for send, _ in self._streams.get(message["type"], []):
+            await send.send(message)
+
+    async def receive(self, message_type: str) -> AsyncIterator[Message]:
+        """Receive messages of a given type"""
+        send, recv = create_memory_object_stream()
+        self._streams.setdefault(message_type, []).append((send, recv))
+        async with recv:
+            async with send:
+                async for message in recv:
+                    yield message
+
+    async def __aenter__(self) -> Messenger:
+        await self._task_group.__aenter__()
+        return self
+
+    async def __aexit__(self, *args) -> None:
+        await self._task_group.__aexit__(*args)
--- a/src/py/reactpy/reactpy/backend/starlette.py
+++ b/src/py/reactpy/reactpy/backend/starlette.py
@@ -3,10 +3,12 @@ from __future__ import annotations
 import asyncio
 import json
 import logging
-from collections.abc import Awaitable
+from collections.abc import AsyncIterator, Awaitable
 from dataclasses import dataclass
 from typing import Any, Callable
 
+from py.reactpy.reactpy.backend.hooks import MessengerContext
+from py.reactpy.reactpy.core.serve import Messenger
 from starlette.applications import Starlette
 from starlette.middleware.cors import CORSMiddleware
 from starlette.requests import Request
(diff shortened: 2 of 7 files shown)


src/py/reactpy/reactpy/backend/messenger.py
```diff
@@ -0,0 +1,61 @@
...
+    def start_consumer(
+        self, message_type: str, consumer: Callable[[Message], Awaitable[None]]
+    ) -> None:
+        """Add a message consumer"""
+
+        async def _consumer() -> None:
+            async for message in self.receive(message_type):
+                self._task_group.start_soon(consumer, message)
+
+        self._task_group.start_soon(_consumer)
```
rmorshea, line 40: These need to return callbacks to stop the consumer/producer - this will work well with `use_effect`.

Archmonger:
Unfortunately, `async for msg in msgr.consume("my-message-type"):` is fundamentally flawed since it relies on our async `use_effect`.

We would need a stateful async paradigm, similar to `Django Channels` consumers.

rmorshea:
At some point we need to figure out async effects. I don't think that should be a blocker here.

Edit: see https://github.com/reactive-python/reactpy/pull/1090

src/py/reactpy/reactpy/backend/messenger.py
```diff
@@ -0,0 +1,61 @@
...
+    async def send(self, message: Message) -> None:
+        """Send a message to all consumers of the message type"""
+        for send, _ in self._streams.get(message["type"], []):
+            await send.send(message)
+
+    async def receive(self, message_type: str) -> AsyncIterator[Message]:
+        """Receive messages of a given type"""
+        send, recv = create_memory_object_stream()
+        self._streams.setdefault(message_type, []).append((send, recv))
+        async with recv:
```
Archmonger, line 51: Would `async with send, receive:` not work here?

src/py/reactpy/reactpy/backend/starlette.py
```diff
@@ -141,6 +143,25 @@ async def model_stream(socket: WebSocket) -> None:
         pathname = pathname[len(options.url_prefix) :] or "/"
         search = socket.scope["query_string"].decode()
 
+        async with Messenger() as msgr:
+            async with Layout(
+                MessengerContext(
+                    ConnectionContext(
```
Archmonger, line 149: We can consider merging everything into a `ReactPyContext`.

src/py/reactpy/reactpy/core/types.py
```diff
@@ -73,6 +73,9 @@ class LayoutType(Protocol[_Render, _Event]):
     async def render(self) -> _Render:
         """Render an update to a view"""
 
+    async def renders(self) -> AsyncIterator[_Render]:
+        """Render a series of updates to a view"""
+
```
Archmonger, line 78: Why not have `AsyncIterator` be an accepted type on `render`, and then handle detection of this within `layout.py`?

Less APIs is generally more intuitive than more.

Archmonger (commented):
I dislike the `start_consumer` and `start_producer` from an "ease of learning" perspective. Following the ReactJS philsophy of diluting the core API to nothing but the most basic interface, we should simplify everything to just `send` and `receive`.

The "easiest to learn" way to advertise this functionality would be as such:

```python
@component
def demo_send_receive():
    @use_messenger
    async def consumer(send, receive):
        while True:
            msg = await receive("my-message-type")
            await send("my-message-type", ...)
```

Just the bare concept of async cancellation, and lack of reliability for other hooks to work in a persistent way makes the user experience really awkward though.

For example, if we force the user to enable `shield_cancel`, we should consider whether `state` and `set_state` will always work as expected in the following scenario:

```python
@component
def demo_send_receive():
    state, set_state = use_state(0)

    @use_messenger
    async def consumer(send, receive):
        while True:
            msg = await msgr.receive("my-message-type")
            print(state)
            set_state(msg + 1)
```

Or alternatively, if we do not enable `shield_cancel` is there potentially a gap in time where the previous `use_effect` has been de-registered, but the new one is not registered yet?

Archmonger:
After today's tag up, two theoretical interfaces could look like this

Notes
- Use messenger will never restart a consumer that is already running
- We should timeout the consumer if it takes too long to teardown
- Async function interface needs a teardown parameter in the hook

## Async Function Interface
```python
from typing import Coroutine
import asyncio
from reactpy import use_state, component, use_effect


def use_messenger(consumer: Coroutine, timeout: int = 10, teardown: Coroutine | None = None):
    ...


@component
def example():
    state, set_state = use_state(0)

    async def consumer_teardown(send, receive):
       ...

    @use_messenger(timeout=20, teardown=consumer_teardown)
    async def consumer(receive, send, cancel: asyncio.Event):
        while True:
            msg = await receive("my-message-type")
            await send("my-message-type", ...)
            print(state)
            set_state(msg + 1)

            if cancel.is_set():
                break
```

## Class Based Interface
```python
from typing import Coroutine
import asyncio
from reactpy import use_state, component, use_effect


def use_messenger(consumer: Coroutine, timeout: int = 10):
    ...


@component
def example():
    state, set_state = use_state(0)

    @use_messenger(timeout=20)
    class Consumer:
        async def __call__(self, receive, send, cancel: asyncio.Event):
            while True:
                msg = await receive("my-message-type")
                await send("my-message-type", ...)
                print(state)
                set_state(msg + 1)

                if cancel.is_set():
                    break

        async def teardown(self):
            ...
```

Archmonger:
Note: We may want to make some considerations around [Django Channels Layers](https://channels.readthedocs.io/en/stable/topics/channel_layers.html) while developing this interface.

It's an existing ASGI messaging paradigm, so taking inspiration from their interface might not be a bad idea.

PR #1085, opened 2023-07-03
Fix issue from #1081

### LINKED ISSUE #1086: Model's JSON Pointer Not Updated
We are not appropriately updating the location of a given component within the overall VDOM when it changes position. [This particular line](https://github.com/reactive-python/reactpy/blob/f065655ae1fc8f93a0ca05769be19e304f607dfa/src/py/reactpy/reactpy/core/layout.py#L492) is at fault - it just blindly copies the path from the old model instead of recomputing that path given the new parent and index (`f"{new_parent.patch_path}/children/{new_index}"`).

### Discussed in https://github.com/reactive-python/reactpy/discussions/1081

<sup>Originally posted by **numpde** July  2, 2023</sup>
I'm confused about what's happening here. Consider this App:

```Python
from reactpy import component, use_state
from reactpy.html import div, button
from reactpy.core.types import State


@component
def Item(item: str, items: State):
    color = use_state(None)

    def deleteme(event):
        items.set_value(lambda items: [i for i in items if (i != item)])

    def colorize(event):
        color.set_value(lambda c: "blue" if not c else None)

    return div(
        {'style': {'background-color': color.value or "transparent", 'padding': "5px"}},
        div(
            button({'onClick': colorize}, f"Color {item}"),
            button({'onClick': deleteme}, f"Delete {item}"),
        ),
    )


@component
def App():
    items = use_state(["A", "B", "C"])

    return div(
        [
            Item(item, items, key=item)
            for item in items.value
        ]
    )

```

I embed it from a Django template. [Here's](https://playcode.io/1521983) the equivalent ReactJS implementation for reference.

Now, when I delete **Item A** then click to color **Item B**, then unexpectedly, **Item C** appears overwritten.

reactpy==1.0.1
reactpy-django==3.2.0
</div>

rmorshea:
<sub style="text-align:right">By submitting this pull request you agree that all contributions to this project are made under the MIT license.</sub>

## Issues

Originally discussed in: https://github.com/reactive-python/reactpy/discussions/1081
Closes: #1086

## Summary

Fix and write a test for the bug.

## Checklist

### DIFF
--- a/src/py/reactpy/reactpy/core/layout.py
+++ b/src/py/reactpy/reactpy/core/layout.py
@@ -489,7 +489,7 @@ def _update_component_model_state(
         index=new_index,
         key=old_model_state.key,
         model=Ref(),  # does not copy the model
-        patch_path=old_model_state.patch_path,
+        patch_path=f"{new_parent.patch_path}/children/{new_index}",
         children_by_key={},
         targets_by_event={},
         life_cycle_state=(
--- a/docs/source/about/changelog.rst
+++ b/docs/source/about/changelog.rst
@@ -31,6 +31,7 @@ Unreleased
 
 - :issue:`930` - better traceback for JSON serialization errors (via :pull:`1008`)
 - :issue:`437` - explain that JS component attributes must be JSON (via :pull:`1008`)
+- :issue:`1086` - fix rendering bug when children change positions (via :pull:`1085`)
 
 
 v1.0.0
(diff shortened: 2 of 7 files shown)


src/py/reactpy/reactpy/core/layout.py
```diff
@@ -489,7 +489,7 @@ def _update_component_model_state(
         index=new_index,
         key=old_model_state.key,
         model=Ref(),  # does not copy the model
-        patch_path=old_model_state.patch_path,
+        patch_path=f"{new_parent.patch_path}/children/{new_index}",
```
rmorshea, line 492: This is the core of the change. Everything else is just tooling to write the test.

PR #1090, opened 2023-07-04
effects with cancel scopes

### LINKED ISSUE #956: Better Async Effect Cleanup

rmorshea:

## Issues

Closes: #956

Presently, there is no way for async effects to shield themselves from cancellation. This is a problem if an effect needs to perform some cleanup before exiting or if the effect is performing an operation that, if it were to stop halfway through, would leave the program in an invalid state.

## Summary

Leverage [`anyio.CancelScope`](https://anyio.readthedocs.io/en/3.x/cancellation.html#shielding)` to allow users to shield important operations from cancellation. Cancellation shielding can be achieved by entering a `CancelScope(shield=True)` async context:

```python
@use_effect
async def my_critical_effect():
    await non_critical_task()

    async with CancelScope(shield=True):
        await critical_task()

    await another_non_critical_task()
```

In the code above, if the effect reaches the `critical_task()`, it will always complete that before exiting, even if the effect has been cancelled by an unmounting component.

The current implementation is not perfect though. We ought to block until all effects associated with the current component have exited. This will require asyncifying some of the underlying rendering logic in `Layout`. This isn't too much work, but it can be left for a follow up PR. In its current state, if a user shielded an effect from cancellation it is technically possible for the lifetime of the cancelled (but currently shielded) effect from overlapping with the lifetime of the effect initiated as part of the next render.

## Checklist

### DIFF
--- a/src/py/reactpy/reactpy/core/hooks.py
+++ b/src/py/reactpy/reactpy/core/hooks.py
@@ -16,6 +16,7 @@ from typing import (
     overload,
 )
 
+from anyio import create_task_group
 from typing_extensions import TypeAlias
 
 from reactpy.config import REACTPY_DEBUG_MODE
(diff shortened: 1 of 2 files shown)


Archmonger:
I think a `prevent_cancel = ...` parameter in `use_effect` might be a convenient way of handling this.

Issue #1075, opened 2023-06-20
Python interfaces for web application APIs


Archmonger:
### Current Situation

Our SPA support isn't really useable until we have Python support for more things.

### Proposed Actions

Read and/or write support for

Application
- Web Manifest
- Service Workers

Storage
- Local Storage / Cookies
- Session Storage
- IndexedDB
- WebSQL
- Cookies
- Interest Groups
- Shared Storage
- Cache Storage

Background Services
- Back/forward cache
- Background Fetch
- Background Sync
- Notifications
- Payment Handler
- Periodic Background Sync
- Push Messaging
- Reporting API

We might also want to expand this to cover client-side features such as webcam access. But anything requiring data streaming (such as webcam access), will require [data serialization](https://github.com/reactive-python/reactpy/issues/574).

Archmonger:
Closing this in favor of #894

PR #1091, opened 2023-07-06
Access to client's localStorage and sessionStorage

williamneto:
## Issues

Discussed in #1083 and #1075 

## Solution

The methods use_local_storage and use_session_storage have been implemented to provide access to these two client storage instances.

## Checklist

### DIFF
--- a/docs/source/about/changelog.rst
+++ b/docs/source/about/changelog.rst
@@ -23,7 +23,8 @@ more info, see the :ref:`Contributor Guide <Creating a Changelog Entry>`.
 Unreleased
 ----------
 
-Nothing yet...
+**Added**
+- :issue:`1075` - Add support to localStorage and sessionStorage access
 
 
 v1.0.2
--- a/src/js/packages/@reactpy/client/src/components.tsx
+++ b/src/js/packages/@reactpy/client/src/components.tsx
@@ -40,6 +40,18 @@ export function Layout(props: { client: ReactPyClient }): JSX.Element {
     [currentModel, props.client],
   );
 
+  useEffect(
+    () =>
+      props.client.onMessage("sync-local-storage", ({ type, storage}) => {
+        for (let itemKey in storage) {
+          window.localStorage.setItem(
+            itemKey,
+            storage[itemKey]
+          )
+        }
+      })
+  )
+
   return (
     <ClientContext.Provider value={props.client}>
       <Element model={currentModel} />
--- a/src/js/packages/@reactpy/client/src/components.tsx
+++ b/src/js/packages/@reactpy/client/src/components.tsx
@@ -52,6 +52,18 @@ export function Layout(props: { client: ReactPyClient }): JSX.Element {
       })
   )
 
+  useEffect(
+    () =>
+      props.client.onMessage("sync-session-storage", ({ type, storage}) => {
+        for (let itemKey in storage) {
+          window.sessionStorage.setItem(
+            itemKey,
+            storage[itemKey]
+          )
+        }
+      })
+  )
+
   return (
     <ClientContext.Provider value={props.client}>
       <Element model={currentModel} />
--- a/src/js/packages/@reactpy/client/src/messages.ts
+++ b/src/js/packages/@reactpy/client/src/messages.ts
@@ -12,6 +12,11 @@ export type LayoutEventMessage = {
   data: any;
 };
 
-export type IncomingMessage = LayoutUpdateMessage;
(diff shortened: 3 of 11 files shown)


olivershetler:
Hello! Just wondering. Why hasn't this push been merged since July? I'm building a frontend service using ReactPy and it's a non-starter to not have access to local storage.

kumaraguru1735:
I am stuck, there is no session_storage or local_storage or cookie like system to work with, In the middle of my dream project
Without any option , how to work with backend api for jwt auth ???
Do something fast...

ipcloudlive:
Even Reflex, Flet having options for LocalStorage, SessionStorage 
This project has released 5 years back, I am still wondering , why its not even having small option and proper docs
This is the reason , no one is using this project....
Even people who start to work on the project will drop their project at the middle, when they realase its a useless module...

Micoder-dev:
Please consider this pull request; it addresses a crucial core feature that we urgently need.

kumaraguru1735:
Hi @Micoder-dev , I also need the same, I donot understand why the developer stand away from this like core option's
**They are developing for their own purpose ?** _What will be on the people's mind, When they on the middle of their dream project ?_
There need to change Reactpy project name to Reactpy | No docs | No options | Nothing
I feel very bad on using this like shit project..

RWallan:
The PR seems to be ok and implements an extremely important feature. Mainly for those who want to put the project into production using Reactpy.
I will be stepping away from the project until such time as a persistent data system exists.
Hope it will be soon.

Archmonger:
@williamneto Hey William. We've been trying to limit features in `reactive-python/reactpy` to things that have direct equivalents to ReactJS core features.

We like the concept of these hooks, but we were wondering if you could fork it into a separate package?

Micoder-dev:
when will I expect the LocalStorage option?

ipcloudlive:
There is no way to use this option ?

Archmonger:
After discussion, we decided we don't want to create ~20 hooks for every single web application API directly within ReactPy core. https://github.com/reactive-python/reactpy/issues/1075

ReactJS philosophy is to focus on primitives within the core library. So following that philosophy, we've landed on a generic `use_script` hook that would work as a better catch-all. https://github.com/reactive-python/reactpy/issues/894

The `use_script` hook will rely on the infrastructure set up within our `use_messenger` PR, so we'll need to get that merged first. https://github.com/reactive-python/reactpy/pull/1084

Hopefully we'll be able to get that pushed through within the next 30-60 days.

williamneto:
As sugested by @Archmonger , the implementation i proposed for local storage is in my fork [williamneto/reactpy](github.com/williamneto/reactpy/). Just merged the latest version of the original into the fork.

There is a usage sample in react/samples

kumaraguru1735:
@williamneto How to build the project?

Archmonger:
@williamneto I wasn't referring to forking the whole project. It would be better make a minimal package on PyPI named something like `reactpy-client-storage`.

That would require PRing ReactPy core to add some third-party extensibility to

- `client/src/components.tsx`
- `client/src/reactpy-client.ts`
- `reactpy/backend`

Also, instead of creating unique properties within the `Connection` object, you can store `local_storage` and `session_storage` within the `Connection.scope` dict.

ipcloudlive:
@Archmonger Better you can create on reactive-python account itself, so new bees will find out easily, and plz solve the issue as soon as possible bcz most people waiting for this option

williamneto:
So @Archmonger , i started studing arround this reactpy-client-storage approach. Realy makes more sense.

But i ended up in the same question presented in #1084 : the necessity to extend the client-server comunications, as by now it only supports  LayoutEvent and LayoutUpdate messages.

After #1084 being merged it will be possible to think about extensions like reactpy-client-storage, reactpy-client-services, etc

kumaraguru1735:
But #1084 is Still in the draft section since July 2023

Archmonger:
@kumaraguru1735 The PR author, Ryan, needed to step away from OSS for a few months. He's back since mid-December and progress will be made on that PR soon.

Archmonger:
For people that need immediate access to these hooks, ~~you can install William's fork via `pip install 'git+https://github.com/williamneto/reactpy.git@main#subdirectory=src/py/reactpy'`~~
**Correction:** It seems that you can't just `git+` install the repo due to our new `hatch` build environment. I'll look into it.

We'll work with him to create a separate package for these hooks once the `use_messenger` primitive hits ReactPy core.

kumaraguru1735:
When Will I expect it ?

PR #1093, opened 2023-07-08
Better Async Effects

### LINKED ISSUE #956: Better Async Effect Cleanup

rmorshea:

## Issues

Closes: #956

## Solution

Async effects now accept a "stop" `Event` that is set when an effect needs to be re-run or a component is unmounting. The next effect will only run when the last effect has exited.

```python
@use_effect
async def my_effect(stop):
    ... # do work
    await stop.wait()
    ... # do cleanup
```

Implementing this same behavior using sync effects is quite challenging:

```python
import threading, asyncio
from reactpy import use_effect, use_ref


def use_async_effect(func):
    thread = use_ref(None)

    @use_effect
    def my_effect():
        if thread.current is not None:
            # wait for last effect to complete
            thread.current.join()

        loop = stop = None
        started = threading.Event()

        def thread_target():
            async def wrapper():
                nonlocal loop, stop
                loop = asyncio.get_running_loop()
                stop = asyncio.Event()
                started.set()
                await func(stop)

            asyncio.run(wrapper())

        # start effect
        thread.current = threading.Thread(target=thread_target, daemon=True)
        thread.current.start()
        started.wait()

        # register stop callback
        return lambda: loop.call_soon_threadsafe(stop.set)
```

To achieve this without requiring an implementation similar to the above, we've asyncified the internals of `Layout`. This has the side-effect of allowing other things to happen while the `Layout` is rendering (e.g. receive events, or allowing the server to respond to other requests). This potentially comes at the cost of rendering speed. For example, if a user is spamming the server with events renders may be interrupted in order to respond to them.

## Checklist

### DIFF
--- a/.github/workflows/.hatch-run.yml
+++ b/.github/workflows/.hatch-run.yml
@@ -16,7 +16,7 @@ on:
       python-version-array:
         required: false
         type: string
-        default: '["3.x"]'
+        default: '["3.11"]'
       node-registry-url:
         required: false
         type: string
--- a/docs/source/about/changelog.rst
+++ b/docs/source/about/changelog.rst
@@ -28,6 +28,14 @@ Unreleased
 - :pull:`1118` - `module_from_template` is broken with a recent release of `requests`
 - :pull:`1131` - `module_from_template` did not work when using Flask backend
 
+**Added**
+
+- :pull:`1093` - Better async effects (see :ref:`Async Effects`)
+- :pull:`1093` - Support concurrent renders - multiple components are now able to render
+  simultaneously. This is a significant change to the underlying rendering logic and
+  should be considered experimental. You can enable this feature by setting
+  ``REACTPY_FEATURE_CONCURRENT_RENDER=1`` when running ReactPy.
+
 
 v1.0.2
 ------
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -322,6 +322,7 @@ intersphinx_mapping = {
     "sanic": ("https://sanic.readthedocs.io/en/latest/", None),
     "tornado": ("https://www.tornadoweb.org/en/stable/", None),
     "flask": ("https://flask.palletsprojects.com/en/1.1.x/", None),
+    "anyio": ("https://anyio.readthedocs.io/en/stable", None),
 }
 
 # -- Options for todo extension ----------------------------------------------
--- a/docs/source/reference/hooks-api.rst
+++ b/docs/source/reference/hooks-api.rst
@@ -89,7 +89,9 @@ Use Effect
 
 .. code-block::
 
-    use_effect(did_render)
+    @use_effect
+    def did_render():
+        ...  # imperative or state mutating logic
 
 The ``use_effect`` hook accepts a function which may be imperative, or mutate state. The
 function will be called immediately after the layout has fully updated.
@@ -117,11 +119,10 @@ then closing a connection:
 
 .. code-block::
 
+    @use_effect
     def establish_connection():
         connection = open_connection()
         return lambda: close_connection(connection)
-
-    use_effect(establish_connection)
 
 The clean-up function will be run before the component is unmounted or, before the next
 effect is triggered when the component re-renders. You can
@@ -141,13 +142,18 @@ example, imagine that we had an effect that connected to a ``url`` state variabl
 
     url, set_url = use_state("https://example.com")
 
+    @use_effect
     def establish_connection():
         connection = open_connection(url)
         return lambda: close_connection(connection)
 
-    use_effect(establish_connection)
-
 Here, a new connection will be established whenever a new ``url`` is set.
+
+.. warning::
+
+    A component will be unable to render until all its outstanding effects have been
+    cleaned up. As such, it's best to keep cleanup logic as simple as possible and/or
+    to impose a time limit.
 
 
 Async Effects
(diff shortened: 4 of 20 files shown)


Archmonger:
This interface should be relatively similar to the [`use_messenger` hook we designed](https://github.com/reactive-python/reactpy/pull/1084#issuecomment-1624765310).

Notes
- We should timeout the effect if it takes too long to teardown
- A 2 second timeout is a good default for most situations.
- Needs a `teardown: Coroutine` parameter in the hook

```python
import asyncio
from reactpy import component, use_effect

@component
def example():
    async def teardown():
       ...

    @use_effect(timeout=2, teardown=teardown)
    async def example_effect(cancel: asyncio.Event):
        while True:
            await something("my-message-type")
            ...

            if cancel.is_set():
                break
```

rmorshea:
> Needs a teardown: Coroutine parameter in the hook

I think this is a slight simplification - teardown should just happen after the stop event is triggered.

```python
@use_effect
async def effect(stop):
    # do effect
    await stop.wait()
    # cleanup
```

> We should timeout the effect if it takes too long to teardown

I thought about this for a bit and realized that a simple `timeout` parameter is ambiguous - does it apply to the effect creation or cleanup. I'd rather not complicate it with more parameters. Adding a timeout is also quite easy:

```python
await asyncio.wait_for(task(), timeout...)
```

Archmonger:
> teardown should just happen after the stop event is triggered.

This can't always occur though. For example, in the case of the user closing their browser window. We need a timeout to cover edge cases like this.

> a simple timeout parameter is ambiguous

We can call it `exit_timeout`, `cancel_timeout`, or `stop_timeout`

> Adding a timeout is also quite easy:
> `await asyncio.wait_for(task(), timeout...)`

What you proposed isn't equivalent to a `stop_timeout`.

rmorshea:
Both comments seem to address the fact that we might want to enforce a timeout when a connection is closed. That seems reasonable, but in that case, it seems better to introduce that timeout here when a `Layout` is exiting:  https://github.com/reactive-python/reactpy/blob/c311345eacea2e22753adebaf2191b516ddee68a/src/py/reactpy/reactpy/core/layout.py#L75

Though, even without this, the user could enforce a "stop timeout" themselves if they wanted:

```python
await wait_for(create_effect(), timeout=...)  # creation timeout
await stop.wait()
await wait_for(cleanup_effect(), timeout=...)  # cleanup timeout
```

rmorshea:
So thinking through our interface some more I'm realizing that:

```python
@use_effect
async def my_effect(stop):
    task = asyncio.create_task(do_something())
    await stop.wait()
    task.cancel()
    await finalize_it()
```

Is not correct since simply cancelling a task does not mean that it will have exited by the time `finalize_effect` starts running. The correct implementation would actually be:

```python
@use_effect
async def my_effect(stop):
    task = asyncio.create_task(do_something())
    await stop.wait()
    task.cancel()
    try:
        await task
    except asyncio.CancelledError:
        pass
    await finalize_it()
```

This is far too complicated for users to understand.

As such, I've done some thinking and realized that in 3.11 `asyncio` got some interesting new features that would allow the above to be implemented with this instead:

```python
@use_effect
async def my_effect(effect):
    async with effect:
        await do_something()
    await finalize_it()
```

The behavior is that awaitables within the `body` will be cancelled if they have not completed by the time the effect needs to stop. The `body` will also pause until the effect needs to be stopped, thus allowing you to perform finalization after exiting the body. If you want to implement a "fire and forget" effect, then you simply ignore the `body` and do not enter it.

The implementation of the `effect` context manager would look something like:

```python
class AsyncEffect:

    _task: asyncio.Task | None = None

    def __init__(self) -> None:
        self._stop = asyncio.Event()
        self._cancel_count = 0

    def stop(self) -> None:
        if self._task is not None:
            self._cancel_task()
        self._stop.set()

    async def __aenter__(self) -> None:
        self._task = asyncio.current_task()
        self._cancel_count = self._task.cancelling()
        if self._stop.is_set():
            self._cancel_task()
        return None

    async def __aexit__(self, exc_type: type[BaseException], *exc: Any) -> Any:
        if exc_type is not asyncio.CancelledError:
            # propagate non-cancellation exceptions
            return None

        if self._task.cancelling() > self._cancel_count:
            # Task has been cancelled by something else - propagate it
            return None

        await self._stop.wait()
        return True

    def _cancel_task(self) -> None:
        assert self._task is not None
        self._task.cancel()
        self._cancel_count += 1
```

Archmonger:
Is there any way for us to use Python 3.11 asyncio features in older versions?

I'm pretty sure the answer is no, and if so what do we want to do in the interim while we wait for 3.11 to become our minimum version?

rmorshea:
I'll have to play around with whether this can be achieved with `anyio`. It may be possible.

Archmonger:
If we're going to introduce async specific `use_effect` parameters, we might want to separate async effects into their own hook.

Such as `use_async_effect`.

rmorshea:
~@Archmonger I'm encountering some [issues with lxml during installation](https://github.com/reactive-python/reactpy/actions/runs/6916649844/job/18816942704?pr=1093). I fixed a similar issue by installing `setuptools` but that [doesn't seem to be helping](https://github.com/reactive-python/reactpy/blob/0319e4db7dd3cbb0d52bd02c6dc4c68cf385c097/src/py/reactpy/pyproject.toml#L38).~

Limiting the default python version to 3.11 seems to have fixed it.

rmorshea:
I'm going to close this and break up the changes into two parts:

1. Concurrent renders
2. Async effects.

PR #1113, opened 2023-07-18
ReactPy ASGI Middleware and standalone ReactPy ASGI App

### LINKED ISSUE #525: Improve Reconnecting WS Configurability
### Current Situation

Currently the configuration for reconnecting WS is fairly limited.

### Proposed Changes

We can potentially add settings for `MAX_TIMEOUT` , `MAX_RETRIES`, and/or `RECONNECT_INTERVAL`. 

Alternatively, allow the user to define his own JavaScript WS reconnection function.

### Implementation Details

> From @rmorshea 
> I'd actually like to modify how reconnections happen so its more configurable. It should accept a callback that receives the last delay, and the total elapsed time as its arguments and returns the next delay.
>
> It should probably be even a bit more general in that, there should be a way to hook into the full lifecycle of the connection:
>
>- before initial connection
>- connected
>- disconnected
>- closed
>- etc...

### LINKED ISSUE #653: Embedding ReactPy via Template Tags

### LINKED ISSUE #1079: Modify `SimpleReactPyClient` to not use pre-defined endpoints
### Current Situation

We currently require a structured/implict routes for `SimpleReactPyClient`.

```js
/*
 * - `_reactpy/stream/${route}${query}`: The websocket endpoint for the stream.
 * - `_reactpy/modules`: The directory containing the dynamically loaded modules.
 * - `_reactpy/assets`: The directory containing the static assets.
*/ 
 ```

### Proposed Actions

Rewrite `SimpleReactPyClient` to allow the user to define their own routes. Will need a prop for each route.

Similar to [`reactpy-django`](https://github.com/reactive-python/reactpy-django/blob/0aaaaac8c44a1eafe34d7be0fb79ad710a0ae2dc/src/js/src/index.js), we need the following customization options:
- `mountId`
   - The string ID of the mountable element
   - `504ef671-57a4-4128-a48d-3ec94c9bfd41`
- `websocketUrl`
   - Path to where the websocket is located at
   - `reactpy/<dotted_path>`
- `jsModulesUrl`
   - Path to where we auto-serve JavaScript modules
   - `reactpy/modules/`
- `reconnectMax`
   - When x many seconds pass, the websocket will give up retries
   - `3600`
- `reconnectInterval`
   - Client will retry to connect to the server every x seconds
   - `5`

Note: We need to start treating the client JS file completely separately from other static files. The user needs can't ever configure their static file dir if we rely on it for core behavior.

### LINKED ISSUE #1110: ReactPy ASGI App and Middleware
### Current Situation

Currently we perform ASGI routing via backend-specific APIs. However, it is much easier to gain broad compatibility via ASGI middleware. Additionally, we should have a "standalone" mode where ReactPy can run in a production configuration without any backend.

I [originally pitched](https://github.com/reactive-python/reactpy/pull/703#issuecomment-1074537270) this concept a long time ago during our development of our `configure()` function.

### Proposed Actions

Create a ReactPy ASGI application and middleware. 

### Interface Design
```python
# This is "standalone mode"
from reactpy.backend import ReactPy

app = ReactPy(my_component)

# This is "middleware mode"
from reactpy.backend import ReactPy
from sanic import Sanic

app = ReactPy(Sanic())
```

### Implementation Draft
```python
import re

from asgiref.compatibility import guarantee_single_callable


class ReactPy:
    def __init__(
        self,
        application=None,
        dispatcher_url="reactpy/stream/${route}${query}",
        modules_url="reactpy/modules",
        static_url="reactpy/assets",
    ) -> None:
        self.user_app = guarantee_single_callable(application)
        self.url_patterns = "|".join((dispatcher_url, modules_url, static_url))

    async def __call__(self, scope, receive, send) -> None:
        """The ASGI callable. This determines whether ReactPy should route the the
        request to ourselves or to the user application."""
        if not self.user_app or re.match(self.url_patterns, scope["path"]):
            await self.reactpy_app(scope, receive, send)
        else:
            await self.user_app(scope, receive, send)

    async def reactpy_app(self, scope, send, receive) -> None:
        """The ASGI application for ReactPy."""
        # This will handle the following: `index.html` view, component dispatcher, web modules, and static files.
```

Archmonger:
## Issues

- fix #1110
- fix #1079
- fix #525
- fix #653

## Solution

Create a ReactPy ASGI application and middleware.

Features of the ASGI application/middleware
- Gain compatibility with every ASGI framework.
- Allows ReactPy to be "self contained" if needed (run within any ASGI webserver without a backend).

For example, this is the interface for ReactPy's standalone mode:

```python
from reactpy import component, html
from reactpy.asgi.standalone import ReactPy

@component
def Example():
    return html.div("Hello World")

app = ReactPy(Example)
# Now you can run `uvicorn example:app --reload` to start ReactPy!
```

## Major Changes
- `ReactPy` class has been created as the standalone API for ReactPy.
- `ReactPyMiddleware` has been created as the version of ReactPy that can wrap other ASGI web frameworks.
- Added a template tag to use alongside our middleware.
- `@reactpy/client` has been rewritten to be more modular. This also helps in preparation for template tags.
- `reactpy.backends.*` is removed.

## Checklist


Archmonger:
The more I develop this, the more I realize it's unwise to finish this PR until the following issues are closed
- #1079 
- #653 

This PR is in a 90% finished state, but I will hold off on finishing it for now and refocus on template tags.

Archmonger:
@rmorshea I will implement things using `starlette` for static files now, but ideally I don't want a web framework to be a dependency for us.

I would feel more comfortable using `whitenoise` as it's a static-file focused library that doesn't have extra stuff. However, this relies on this PR getting merged:
- https://github.com/evansd/whitenoise/pull/359

Archmonger:
Note to self:
It seems logical to mandate that ReactPy standalone's root component must be a `html.html` node. By forcing the user to template the whole page, this would eliminate the need for a janky `Options(head=...)` parameter.

```python
html.html(
    {"lang": "en"},
    html.head(
        ...,
    ),
    html.body(
        ...,
    ),
)
```

It also seems logical to have `ReactPyStandalone` and `ReactPyMiddleware` to be discrete classes.

EDIT: Not realistic since the root object needs to be a component, not VDOM.

Archmonger:
The `whitenoise` maintainers have largely been MIA, so I went ahead and forked it into my own package [ServeStatic](https://github.com/Archmonger/ServeStatic).

This PR is no longer blocked.

PR #1164, opened 2023-11-20
Add <marquee> element

Minejerik:
This commit simply adds the option to use the marquee element in reactpy. While yes, the marquee element has been deprecated, I think that it would be best to allow web developers the freedom to use this element even when it is deprecated. Also the marquee element is still usable in all major browsers.

## Issues

The issue was that I was unable to use the marquee element in a project I was working on.

## Solution

I simply added the marquee element in html.py

## Checklist

### DIFF
--- a/src/py/reactpy/reactpy/html.py
+++ b/src/py/reactpy/reactpy/html.py
@@ -281,6 +281,7 @@ __all__ = (
     "var",
     "video",
     "wbr",
+    "marquee",
 )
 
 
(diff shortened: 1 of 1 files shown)


rmorshea:
Looks like `<marquee>` is [deprecated according to Mozilla](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/marquee). Given that, it doesn't seem like we should add this to ReactPy. If needed, you can always create this using `reactpy.vdom("marquee", ...)`. Thanks for the contribution though!

Issue #1136, opened 2023-08-28
Allow threading for sync `use_effect` calls


Archmonger:
### Current Situation

Users frequently run into unexpected webserver "freezing" when using sync effects due to Python's single-threaded nature. This is seemingly caused by the ASGI websocket consumer being blocked from receiving/sending data during hook operations. This behavior doesn't seem to match up with equivalent JavaScript server side frameworks.

### Proposed Actions

By default we should run all sync effects in a `ThreadPoolExecutor(max_workers=1)`. This effectively runs all sync effects in a background thread. Additionally, by running all effects in the same worker thread (`max_workers=1`) this implementation is semi-threadsafe (as long as the user didn't start an operation outside of his `use_effect` call. For operations that must run on the main thread, we need a `@use_effect(threading: bool = ...)` parameter to switch back to legacy behavior.

We could also consider a `reactpy.config.USE_EFFECT_WORKERS=...` setting to allow users to modify the value of `max_workers`.

PR #1165, opened 2023-11-28
Concurrent Renders

### LINKED ISSUE #557: Improve ReactPy Performance
# Proposed Changes

Find the hot spots in ReactPy's code-base in order to identify where attention should be focused. Once we know this, we will be able to focus our efforts on performance improvements.

# Implementation Details

### Analysis
Use [Yappi](https://github.com/sumerc/yappi) to profile ReactPy dispatchers and determine what is taking the most significant amounts of time.

We will also need to figure out how to test websocket concurrency. It seems like there are very few tools for this, especially for our applications. We're probably going to have to create out own benchmarking tools. Here's some resources we might want to read through later.
- https://healeycodes.com/websocket-benchmarker
- https://kemalcr.com/blog/2016/11/13/benchmarking-and-scaling-websockets-handling-60000-concurrent-connections/
- https://ma.ttias.be/benchmarking-websocket-server-performance-with-artillery/

### Sync and Async

Currently, ReactPy relies on synchronous code to properly queue renders. Sync code is known to blocks the asyncio event loop, which causes concurrency issues.

See section "Two Worlds of Python" for why this is a problem:
https://arunrocks.com/a-guide-to-asgi-in-django-30-and-its-performance/

All synchronous functions should be converted to async within ReactPy core. They could alternatively be threadpooled. See reactive-python/reactpy-django#31 for the original discussion on this.

### Extreme Countermeasures
If optimizations can't be suitably performed within Python, consider writing C/C++ code exposed via Python APIs that performs the same functionality as the non-performant parts of IDOM.

rmorshea:

## Issues

Currently ReactPy renders components sequentially. This means that larger renders can block smaller updates from taking place which impacts responsiveness of the page. Sequential renders are also problematic for async effects since effects which are slow to start and stop can similarly impact the responsiveness of a page.

## Solution

This change now allow renders to take place concurrently. To keep things simple, no effort is made to deduplicate renders. For example, if parent and child components are scheduled to render at the same time, both scheduled renders will take place even though a single render of the parent component would be sufficient to update the view.

Concurrent renders are achieved by "asyncifying" the internals of the `Layout` class. Additionally all scheduled renders immediately result in a "render task" being started. To avoid rendering the same component at the same time, we introduce a semaphore inside `LifeCycleHook`s that must be acquired before a component is rendered.

_Note: this contains a subset of the changes originally created in https://github.com/reactive-python/reactpy/pull/1093_

## Checklist


docs/source/about/changelog.rst
```diff
@@ -28,6 +28,13 @@ Unreleased
 - :pull:`1118` - `module_from_template` is broken with a recent release of `requests`
 - :pull:`1131` - `module_from_template` did not work when using Flask backend
 
+**Added**
+
+- :pull:`1165` - Allow concurrent renders of distinct components - enable this
+  experimental feature by setting `REACTPY_FEATURE_CONCURRENT_RENDERING=true`.
```
Archmonger: Is there any point to the having `_FEATURE_` within this env name?

src/py/reactpy/reactpy/core/_life_cycle_hook.py
```diff
@@ -0,0 +1,220 @@
...
+def current_hook() -> LifeCycleHook:
+    """Get the current :class:`LifeCycleHook`"""
+    hook_stack = _HOOK_STATE.get()
+    if not hook_stack:
+        msg = "No life cycle hook is active. Are you rendering in a layout?"
+        raise RuntimeError(msg)
+    return hook_stack[-1]
+
+
+class LifeCycleHook:
```
Archmonger, line 34: A few methods in this class are missing docstrings.

src/py/reactpy/reactpy/core/_life_cycle_hook.py
```diff
@@ -0,0 +1,220 @@
...
+    """Defines the life cycle of a layout component.
```
Archmonger: Life cycle is _technically_ jargon, especially to the uninitiated contributor. Would be a good idea to add a sentence after this first line to explain what a life cycle is.

src/py/reactpy/reactpy/core/layout.py
```diff
@@ -416,10 +455,25 @@ def _unmount_model_states(self, old_states: list[_ModelState]) -> None:
             if model_state.is_component_state:
                 life_cycle_state = model_state.life_cycle_state
                 del self._model_states_by_life_cycle_state_id[life_cycle_state.id]
-                life_cycle_state.hook.affect_component_will_unmount()
+                await life_cycle_state.hook.affect_component_will_unmount()
```
Archmonger, line 458: `affect_component_will_unmount` has always been a bit unclear naming to me. I think it's the word `affect` that has always thrown me off.

docs/source/about/changelog.rst
```diff
@@ -28,6 +28,13 @@ Unreleased
 - :pull:`1118` - `module_from_template` is broken with a recent release of `requests`
 - :pull:`1131` - `module_from_template` did not work when using Flask backend
 
+**Added**
+
+- :pull:`1165` - Allow concurrent renders of distinct components - enable this
```
Archmonger: Might want to add a note that this does **not** allow one component to concurrently render itself multiple times.

I'm sure someone will ask at some point.

Archmonger (commented):
Here's a couple of side comments that aren't related to specific LOCs.

1. Should we start considering allowing async components? I think they'd lead to some fairly clever design patterns. But, those are completely new design patterns that break ReactPy out of the "ReactJS mold".
2. Should we use this PR to add #1136?
3. After merging, we will need a new issue for an deduplication algorithm.
    - If a `parent` is already in the render queue, don't append a new child render.
    - If a `child` is already in the render queue, remove it from the queue.
4. I set this PR to close #557 since JSON patch is the only thing remaining. We've discussed that is a separate can of worms. Patch might not even be worth the computation time, or at the very least it should be toggleable.
5. We will need to create an issue to remind ourselves to change this default from `False` to `True` in v2. Or at least evaluate whether we should do that,

docs/source/about/changelog.rst
rmorshea: Not really. Can definitely remove that.

src/py/reactpy/reactpy/core/layout.py
rmorshea: It's definitely an odd naming convention. I think I was trying to indicate what the life cycle methods were when I originally did it. Definitely something that could be changed in a re-write of `Layout`'s internals.

docs/source/about/changelog.rst
Archmonger: Also we need to decide if `distinct` or `discrete` is the proper word to describe exclusive component trees.

We just need to keep our verbiage consistent. I've been using the word `discrete` (not to say that is correct).

src/py/reactpy/reactpy/core/_life_cycle_hook.py
```diff
@@ -0,0 +1,220 @@
...
+        self._rendered_atleast_once = True
+        self._current_state_index = 0
+        self._render_access.release()
+        del self.component
+
+    async def affect_layout_did_render(self) -> None:
+        """The layout completed a render"""
+        stop = Event()
+        self._effect_stops.append(stop)
+        self._effect_tasks.extend(create_task(e(stop)) for e in self._effect_funcs)
```
rmorshea, line 215: I think this needs to be moved into `affect_component_did_render`

docs/source/about/changelog.rst
rmorshea: It probably also makes sense to be more explicit here and say "(distinct | discrete) component trees" rather than just "distinct components".

rmorshea:
I'll be able to finish this off over the weekend.

docs/source/about/changelog.rst
Archmonger: Maybe `REACTPY_ASYNC_RENDERING` is a better name?

src/py/reactpy/reactpy/core/_life_cycle_hook.py
```diff
@@ -0,0 +1,245 @@
...
+                # render the component
+                ...
+
+                # the component may access the current hook
+                assert current_hook() is hook
+
+                # and save state or add effects
+                current_hook().use_state(lambda: ...)
+
+                async def my_effect(stop_event):
```
Archmonger, line 84: We need docs for how to use `stop_event`

rmorshea: Described [here](https://github.com/reactive-python/reactpy/blob/cd9f5273d4696c745be678cf2aaa930679984481/src/py/reactpy/reactpy/core/_life_cycle_hook.py#L171-L174).

Archmonger: Docstrings do not suffice. It needs to be more user-facing than that, specifically within the "async effects" section in [`hooks-api.rst`](https://reactpy.dev/docs/reference/hooks-api.html#async-effects)

rmorshea: The `add_effect` method isn't intended to be user facing - it's more of a low level API that `use_effect` relies on (hence this all being in `_life_cycle_hook.py`. The user-facing async effect API is going to come in a follow up once this lands.

Archmonger:
There's still a couple of comments on [my original review](https://github.com/reactive-python/reactpy/pull/1165#pullrequestreview-1752061860) that need discussion.

PR #1169, opened 2023-12-09
better async effects

### LINKED ISSUE #956: Better Async Effect Cleanup

rmorshea:

## Issues

Closes: https://github.com/reactive-python/reactpy/issues/956

## Solution

Effects can now be defines as sync or async generators (more info in [the docs](https://github.com/rmorshea/reactpy/blob/af162be9960751c2dda9be7e7b97d367597df111/docs/source/reference/hooks-api.rst?plain=1#L87-L241)):

```python
@use_effect
def my_effect():
    conn = open_connection()
    try:
        # do something with the connection
        yield
    finally:
        conn.close()

@use_effect
async def my_effect():
    conn = await open_connection()
    try:
        # do something with the connection
        yield
    finally:
        await conn.close()
```

_Note: this contains a subset of the changes originally created in https://github.com/reactive-python/reactpy/pull/1093_

## Checklist

### DIFF
--- a/docs/source/reference/hooks-api.rst
+++ b/docs/source/reference/hooks-api.rst
@@ -89,7 +89,9 @@ Use Effect
 
 .. code-block::
 
-    use_effect(did_render)
+    @use_effect
+    def did_render():
+        ...  # imperative or state mutating logic
 
 The ``use_effect`` hook accepts a function which may be imperative, or mutate state. The
 function will be called immediately after the layout has fully updated.
@@ -111,22 +113,45 @@ Cleaning Up Effects
 ...................
 
 If the effect you wish to enact creates resources, you'll probably need to clean them
-up. In such cases you may simply return a function that addresses this from the
-``did_render`` function which created the resource. Consider the case of opening and
-then closing a connection:
+up. In this case you can write a generator function that will ``yield`` until your
+cleanup logic needs to run. Similarly to a
+`context manager <https://realpython.com/python-with-statement/>`__, you'll also need to
+use a ``try/finally`` block to ensure that the cleanup logic always runs:
 
 .. code-block::
 
+    @use_effect
     def establish_connection():
-        connection = open_connection()
-        return lambda: close_connection(connection)
+        conn = open_connection()
+        try:
+            # do something with the connection
+            yield
+        finally:
+            conn.close()
 
-    use_effect(establish_connection)
+.. warning::
+    
+    If you never ``yield`` control back to ReactPy, then the component will never
+    re-render and the effect will never be cleaned up. This is a common mistake when
+    using ``use_effect`` for the first time.
 
-The clean-up function will be run before the component is unmounted or, before the next
-effect is triggered when the component re-renders. You can
+The clean-up logic inside the ``finally`` block will be run before the component is
+unmounted or, before the next effect is triggered when the component re-renders. You can
 :ref:`conditionally fire events <Conditional Effects>` to avoid triggering them each
 time a component renders.
+
+Alternatively to the generator style of cleanup, you can return a cleanup function from
+your effect function. As with the generator style, the cleanup function will be run
+before the component is unmounted or before the next effect is triggered when the
+component re-renders:
+
+.. code-block::
+
+    @use_effect
+    def establish_connection():
+        conn = open_connection()
+        # do something with the connection
+        return conn.close
 
 
 Conditional Effects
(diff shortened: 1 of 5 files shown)


Archmonger:
Will review on Monday. Not home at the moment.

docs/source/reference/hooks-api.rst
```diff
@@ -111,23 +113,46 @@ Cleaning Up Effects
...
+            yield
+        finally:
+            conn.close()
 
-    use_effect(establish_connection)
+.. warning::
+    
+    If you never ``yield`` control back to ReactPy, then the component will never
+    re-render and the effect will never be cleaned up. This is a common mistake when
+    using ``use_effect`` for the first time.
```
Archmonger, line 136: I don't like the idea of having no safeguard against this. An interface that doesn't protect against critical breaking behavior is usually the wrong interface.

Looks like we can leverage [`inspect.isgeneratorfunction`](https://docs.python.org/3/library/inspect.html#inspect.isgeneratorfunction) to auto-detect whether something yields.

docs/source/reference/hooks-api.rst
```diff
@@ -141,40 +166,77 @@ example, imagine that we had an effect that connected to a ``url`` state variabl
...
         connection = open_connection(url)
         return lambda: close_connection(connection)
 
-    use_effect(establish_connection)
-
 Here, a new connection will be established whenever a new ``url`` is set.
 
+.. warning::
+
+    A component will be unable to render until all its outstanding effects have been
```
Archmonger, line 178: render -> re-render

docs/source/reference/hooks-api.rst
```diff
@@ -141,40 +166,77 @@ example, imagine that we had an effect that connected to a ``url`` state variabl
...
+
 
 Async Effects
 .............
 
 A behavior unique to ReactPy's implementation of ``use_effect`` is that it natively
-supports ``async`` functions:
+supports ``async`` effects. Async effect functions may either be an async function
+or an async generator. If your effect doesn't need to do any cleanup, then you can
+simply write an async function.
```
Archmonger, line 189: `then you can simply write` -> `then you should use`

src/py/reactpy/reactpy/core/_life_cycle_hook.py
```diff
@@ -212,7 +212,12 @@ async def affect_layout_did_render(self) -> None:
         """The layout completed a render"""
         stop = Event()
         self._effect_stops.append(stop)
-        self._effect_tasks.extend(create_task(e(stop)) for e in self._effect_funcs)
+        for effect_func in self._effect_funcs:
+            effect_task = create_task(effect_func(stop))
+            # potential memory leak if the task doesn't remove itself when complete
```
Archmonger, line 217: This comment needs a bit more description.

Is the memory leak due to our internal design, or due to user failure? Is the next LOC designed to prevent this issue, or is it the cause of this issue?

src/py/reactpy/reactpy/core/_life_cycle_hook.py
```diff
@@ -212,7 +212,12 @@ async def affect_layout_did_render(self) -> None:
...
+            effect_task.add_done_callback(lambda t: self._effect_tasks.remove(t))
+            self._effect_tasks.add(effect_task)
+        self._effect_tasks.update()
```
Archmonger, line 220: Is there any reason to call `update` here?

src/py/reactpy/reactpy/core/hooks.py
```diff
@@ -130,48 +141,50 @@ def use_effect(
...
-            if last_clean_callback.current is not None:
-                last_clean_callback.current()
-                last_clean_callback.current = None
-            clean = last_clean_callback.current = sync_function()
+    last_effect: Ref[tuple[Event, Task[None]] | None] = use_ref(None)
+
+    def add_effect(function: _Effect) -> None:
+        effect_func = _cast_async_effect(function)
+
+        async def run_effect(stop: Event) -> None:
```
Archmonger, line 158: Does this function really need to be dynamically generated within `add_effect`? It would be more memory efficient to move `run_effect` outside of `add_effect`, and pass in all needed values via args.

Dynamically defined functions are generally a bad idea for any long-running Python code due to memory usage.

src/py/reactpy/reactpy/core/hooks.py
```diff
@@ -180,6 +193,53 @@ async def effect(stop: asyncio.Event) -> None:
...
+            cleanup = await async_function_effect()
+            try:
+                yield
+            finally:
+                if cleanup is not None:
+                    warn(
+                        f"Async effect {async_function_effect} returned a cleanup "
+                        f"function - use an async generator instead by yielding inside "
+                        "a try/finally block. This will be an error in a future "
+                        "version of ReactPy.",
```
Archmonger, line 212: This needs to be mentioned in the changelog.

docs/source/reference/hooks-api.rst
```diff
@@ -111,23 +113,46 @@ Cleaning Up Effects
...
-effect is triggered when the component re-renders. You can
+The clean-up logic inside the ``finally`` block will be run before the component is
+unmounted or, before the next effect is triggered when the component re-renders. You can
 :ref:`conditionally fire events <Conditional Effects>` to avoid triggering them each
 time a component renders.
 
+Alternatively to the generator style of cleanup, you can return a cleanup function from
+your effect function. As with the generator style, the cleanup function will be run
+before the component is unmounted or before the next effect is triggered when the
+component re-renders:
```
Archmonger, line 146: The fact that returning cleanup functions will continue being allowed on sync effects (but not async effects) really makes me think we should commit to a `use_async_effect` hook.

It feels confusing to mandate two completely separate user interfaces within the same hook.

Archmonger (changes requested):
I like the flexibility of generator approach a lot more than our previously proposed `stop_event` interface.

But I do still think [safeguards](https://github.com/reactive-python/reactpy/pull/1093#issuecomment-1627597603) (`timeout`) need to be put in place for programmer error. We can't trust everyone to write perfect code, so we need to detect, warn, then mitigate programming flaws.

On a different note, our async function interface has deviated wildly from the sync equivalent. It's starting to become clear to me we shouldn't overload `use_effect` for async support any longer. We need to migrate towards a `use_async_effect` hook.

rmorshea:
I think we can have `use_async_effect` with a timeout parameter, but something I realized while working on this is that I don't think we can have a default. Having a default timeout trades concerns about responsiveness and hangs for memory leaks and unexpected behavior due to premature timeouts.

Archmonger:
Django has a similar timeout design for views. They have an arbitrary default of 20 seconds.
ASGI webservers also have similar timeouts, specifically related to `keep-alive` requests. Usually default of 5 seconds.

We just have to pick a "good enough" default, and trust the user to change it when needed.

Archmonger:
On that note, maybe `teardown_timeout` is a more appropriate name.

rmorshea:
> We just have to... trust the user to change it when needed.

Ultimately we have to trust the user to know what they're doing in some regard. Either we trust them to understand that if their effects don't end in a timely manner it will impact the responsiveness of components, or we trust that they understand how premature timeouts may cause memory leaks and unexpected behavior when resources aren't cleaned up. My intuition is that debugging issues with responsiveness will be easier than trying to find the source of memory leaks and orphaned resources.

Archmonger:
In the scenario where we auto-teardown, we've caught the error and can log the LOC. But in the scenario of user failure, it's a complete freeze with no user information.

Maybe we can default it to something large like 20 seconds? At that point we'd be 98% confident it's a programming failure and worth an `ERROR` log.

Archmonger:
I'm realizing sync effects can be generators too.

Maybe we can stick with one `use_effect` but remove the ability to return a cancel function?

Archmonger:
In terms of whether we diverge to creating `use_async_effect`, this issue is also relevant: https://github.com/reactive-python/reactpy/issues/1136

In my mind, threading should be unique to sync calls.

Archmonger:
Someone already thought of these issues and developed something relatively clean in the ReactJS space.

Ref: https://marmelab.com/blog/2023/01/11/use-async-effect-react.html#introducing-the-code-classlanguage-textuseasynceffectcode-hook

Archmonger:
Closing this since the PR is stale.

I will be spending some time to rethink this. This PR does significantly deviates us from ReactJS syntax, so perhaps it needs more thought.

PR #1171, opened 2023-12-10
Skip rendering None in all situations

### LINKED ISSUE #1025: Fix `None` within `reactpy.html`
### Current Situation

Currently `None` within `reactpy.html` elements get rendered as the text literal `None`, rather than doing nothing.

Tangentially related:
- https://github.com/reactive-python/reactpy/issues/538
- https://github.com/reactive-python/reactpy/issues/459

### Proposed Actions

We should treat `None` as a positional placeholder that does nothing. We need to make sure that it doesn't mess up how we auto-generate positional keys.

If we want to handle this server-side, we can substitute it with `html._` as a quick-n-easy hack.

rmorshea:

## Issues

Currently the value `None` will render as the string `"None"` when used as a child in VDOM elements. However, when returned by a component it will not render. This inconsistency is confusing for users.

fix #1025

## Solution

We should skip rendering `None` in all cases.

## Checklist

- [X] The `changelog.rst` has been updated with any significant changes.

### DIFF
--- a/docs/source/about/changelog.rst
+++ b/docs/source/about/changelog.rst
@@ -41,10 +41,9 @@ Unreleased
   the string ``"None"``. Now ``None`` will not render at all. This is consistent with
   how ``None`` is handled when returned from components. It also makes it easier to
   conditionally render elements. For example, previously you would have needed to use a
-  fragment to conditionally render an element (e.g.
-  ``something if condition else html._()``). Now you can write:
-  ``something if condition else None``. The latter now has the minor performance
-  advantage of not needing to create and render a fragment.
+  fragment to conditionally render an element by writing
+  ``something if condition else html._()``. Now you can simply write
+  ``something if condition else None``.
 
 
 v1.0.2
--- a/docs/source/about/changelog.rst
+++ b/docs/source/about/changelog.rst
@@ -45,6 +45,12 @@ Unreleased
   ``something if condition else html._()``. Now you can simply write
   ``something if condition else None``.
 
+**Deprecated**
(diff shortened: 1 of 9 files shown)


docs/source/about/changelog.rst
```diff
@@ -35,6 +35,18 @@ Unreleased
...
+**Changed**
+
+- :pull:`1171` - Previously ``None``, when present in an HTML element, would render as
+  the string ``"None"``. Now ``None`` will not render at all. This is consistent with
+  how ``None`` is handled when returned from components. It also makes it easier to
+  conditionally render elements. For example, previously you would have needed to use a
+  fragment to conditionally render an element (e.g.
+  ``something if condition else html._()``). Now you can write:
+  ``something if condition else None``. The latter now has the minor performance
+  advantage of not needing to create and render a fragment.
```
Archmonger: The sentence about a performance advantage is a bit TMI.

src/py/reactpy/reactpy/core/layout.py
```diff
@@ -332,7 +331,7 @@ async def _render_model_children(
         child_type_key_tuples = list(_process_child_type_and_key(raw_children))
 
         new_keys = {item[2] for item in child_type_key_tuples}
-        if len(new_keys) != len(raw_children):
+        if len(new_keys) != len(child_type_key_tuples):
```
Archmonger: Not a fan of the variable name `child_type_key_tuples`

Archmonger (dismissed):

PR #1193, opened 2024-02-10
fix black lint

rmorshea:

## Issues

Lint is broken right now due to changes in Black

## Solution

Pin black + other lint tools to avoid in future

## Checklist


Archmonger:
Should we consider pulling black out of the repo and instead relying on just `ruff`'s format checker?

Issue #993, opened 2023-05-24
Use props `dict` on components, rather than kwargs.


Archmonger:
### Current Situation

There's been a lot of growing pains in the fact that components have a significantly different interface than `reactpy.html` elements.

Developing certain parts of `reactpy` core and/or `reactpy` related utilities has been significantly more difficult due to this interface mismatch.

### Proposed Actions

The new component interface should match the [VDOM interface](https://github.com/reactive-python/reactpy/blob/7428340b44c8db5728fcff3b36908735ecfd4ca3/src/py/reactpy/reactpy/core/types.py#L198-L213), and will look like this:

```python
# `key=...` should be declared within props, identical to the `reactpy.html` interface.
my_component({"prop_1": True, "key": 123}, child_1, child_2)
```

As a part of this change, we should rename `VdomAttributes` to `Props`.

### Do we retain args/kwargs?

We can handle `@component` function definitions a few ways. Some of these ways feel more similar to ReactJS, but have problems with Python type hints. 

1) Force all users to only use kwargs
   - Retain type hints via [PEP692 ](https://peps.python.org/pep-0692/)
2) Allow args/kwargs in function defs. Let Python unpacking operator handle the rest.
   - This would break type hints, since args can't be defined within a TypedDict
3) Force all users to define a `props: dict` as their first parameter
   - However, forcing users to manually define a TypedDict for every component would be extremely inconvenient.

I'm personally leaning towards option 1.

rmorshea:
More precisely the interface should conform to [this protocol](https://github.com/reactive-python/reactpy/blob/main/src/reactpy/core/types.py#L198-L213).

Archmonger:
We can potentially retain type hints using this PEP https://peps.python.org/pep-0692/

Archmonger:
I spent more thought on this issue while rewriting the docs.

@rmorshea We need to decide what we will do with the user's component function defs. See my updates above.

rmorshea:
Note: we'll have to figure out how this will play work if the component function is a method.

Archmonger:
I did some prototyping, and learned it's currently not possible to use `ParamSpec` to provide type hints to a dictionary.

Given that components are functions, it feels natural for users to call them via the standard "python function interface" (AKA arguments within parenthesis)

[...]

PR #1199, opened 2024-02-20
Fix component decorator eating static type hints

JamesHutchison:

## Issues

The `@component` decorator eats type hints so you can't easily see the parameters to your component (just shows up as `...`)

## Solution

This updates the type hinting so that component returns the same things its decorating

## Checklist

before-after-component-type-fix

The top screenshot is the current behavior. The bottom is the updated behavior. Note that the top `Component` is reactpy's while the bottom `Component` is my own.

### DIFF
--- a/src/py/reactpy/reactpy/core/component.py
+++ b/src/py/reactpy/reactpy/core/component.py
@@ -2,14 +2,15 @@ from __future__ import annotations
 
 import inspect
 from functools import wraps
-from typing import Any, Callable
+from typing import Any, Callable, TypeVar
 
 from reactpy.core.types import ComponentType, VdomDict
 
+T = TypeVar("T", bound=Callable[..., ComponentType | VdomDict | str | None])
 
 def component(
-    function: Callable[..., ComponentType | VdomDict | str | None]
-) -> Callable[..., Component]:
+    function: T
+) -> T:
     """A decorator for defining a new component.
 
     Parameters:
--- a/src/py/reactpy/reactpy/core/component.py
+++ b/src/py/reactpy/reactpy/core/component.py
@@ -2,15 +2,17 @@ from __future__ import annotations
 
 import inspect
 from functools import wraps
-from typing import Any, Callable, TypeVar
+from typing import Any, Callable, TypeVar, ParamSpec
 
 from reactpy.core.types import ComponentType, VdomDict
 
-T = TypeVar("T", bound=Callable[..., ComponentType | VdomDict | str | None])
+T = TypeVar("T", bound=ComponentType | VdomDict | str | None)
+P = ParamSpec("P")
+
 
 def component(
-    function: T
-) -> T:
+    function: Callable[P, T],
+) -> Callable[P, Component]:
     """A decorator for defining a new component.
 
     Parameters:
@@ -26,7 +28,7 @@ def component(
         raise TypeError(msg)
 
     @wraps(function)
-    def constructor(*args: Any, key: Any | None = None, **kwargs: Any) -> Component:
+    def constructor(*args: P.args, key: Any | None = None, **kwargs: P.kwargs) -> Component:
         return Component(function, key, args, kwargs, sig)
 
     return constructor


src/py/reactpy/reactpy/core/component.py
```diff
@@ -2,14 +2,15 @@
...
-from typing import Any, Callable
+from typing import Any, Callable, TypeVar
 
 from reactpy.core.types import ComponentType, VdomDict
 
+T = TypeVar("T", bound=Callable[..., ComponentType | VdomDict | str | None])
 
 def component(
-    function: Callable[..., ComponentType | VdomDict | str | None]
-) -> Callable[..., Component]:
```
JamesHutchison, line 12: I'm not sure the significance of going from `ComponentType | VdomDict | str | None` to `Component`. In my own code, I have a `Component` type which is more or less what the function is. I'll play with this on my environment and come back with changes or an A-OK to whether I am finding issues.

Archmonger:
It's been a while since these type hints have been broken, but if I recall they broke due to some type hinting restrictions that cropped up.

The issue was that a component should contain all of the args and kwargs of their parent (as you mention) but they also need to contain one additional kwarg `key=...`.

At the time there was no solution, but I think the concatenate function in the newer Python typing libraries might be able to be jerry-rigged to help here?

src/py/reactpy/reactpy/core/component.py
Archmonger: Writing this response based on my memory but...

You can think of ComponentType as the generic for component. Because it's a protocol it technically allows for external libraries to define their own `Component` class without having to inherit from ours.

JamesHutchison:
This is... closer?

```python

from typing import Any, Callable, Concatenate, TypeVar, ParamSpec

from reactpy.core.types import ComponentType, VdomDict

T = TypeVar("T", bound=ComponentType | VdomDict | str | None)
P = ParamSpec("P")
Q = ParamSpec("Q")


def _key_param_spec(*, key: Any | None = None) -> None:
    pass


def component(
    function: Callable[P, T], __type_helper: Callable[Q, None]=_key_param_spec
) -> Callable[Concatenate[P, Q], Component]:
    """A decorator for defining a new component.

    Parameters:
        function: The component's :meth:`reactpy.core.proto.ComponentType.render` function.
    """
    sig = inspect.signature(function)

    if "key" in sig.parameters and sig.parameters["key"].kind in (
        inspect.Parameter.KEYWORD_ONLY,
        inspect.Parameter.POSITIONAL_OR_KEYWORD,
    ):
        msg = f"Component render function {function} uses reserved parameter 'key'"
        raise TypeError(msg)

    @wraps(function)
    def constructor(*args: P.args, key: Any | None = None, **kwargs: P.kwargs) -> Component:
        return Component(function, key, args, kwargs, sig)

    return constructor
```

I haven't tested this well. It may look gross in PyCharm or mypy or pyright might complain. I don't have mypy set up in my project, yet.
component-refactor

On a side note, shouldn't `key` be a `str | None`?

Archmonger:
No, `key` can be any serializable value. That includes integers and some objects like `UUID`.

rmorshea:
> concatenate function in the newer Python typing libraries might be able to be jerry-rigged to help here?

Unfortunately `Concatenate` does not help. The PEP [explicitly rejects concatenating keyword arguments](https://peps.python.org/pep-0612/#concatenating-keyword-parameters).

Archmonger:
Played around with this for a bit... I'm not seeing a way to fix this without changing our component interface.

Kind of upsetting that there are so many Python type hinting limitations.

There is a realistic future where pyright/mypy could infer types from an implicit return like this...

```python
def component(function: Callable[P, T]):
    sig = inspect.signature(function)

    if "key" in sig.parameters and sig.parameters["key"].kind in (
        inspect.Parameter.KEYWORD_ONLY,
        inspect.Parameter.POSITIONAL_OR_KEYWORD,
    ):
        msg = f"Component render function {function} uses reserved parameter 'key'"
        raise TypeError(msg)

    def constructor(
        *args: P.args, key: Any | None = None, **kwargs: P.kwargs
    ) -> Component:
        return Component(function, key, args, kwargs, sig)

    return constructor
```

But unfortunately, seems like type checkers just give up implicit parsing as soon as `key` (or any new kwarg) is added to the function definition.

JamesHutchison:
What I have provides a clean view of the arguments. It doesn't include `key` in the argument hint, nor does it flag wrong arguments, but that feels better than just seeing `...` in the IDE.
better-than-nothing

It could be stricter settings might get flagged. I know pyright used by VSCode by default doesn't flag many things.

Archmonger:
Mypy flags `key` as an invalid keyword with the latest implementation in your branch.

image

Archmonger:
I've [created a post](https://discuss.python.org/t/type-hints-for-kwarg-overrides/47196) over at the Python typing forums to see what the sentiment is around fixing this type hinting limitation.

Archmonger:
Marking this as a draft since it's broken with most linters.

Archmonger:
If we shove `key` into the props dictionary `{"onClick": ..., "key"=...}` then this solution could move forward.

Archmonger:
Closing in favor of https://github.com/reactive-python/reactpy/issues/993 which will need to be drafted in the future.

Issue #1198, opened 2024-02-19
`use_effect`'s unmount method is not always called with dynamically rendered children


Archmonger:
### Current Situation

This issue was first discovered while trying to develop https://github.com/reactive-python/reactpy-django/pull/226

I created a component that simulates a `list` of conditionally rendered children. Each child in the list contains a unique `key`. However, the `use_effect` unmount method does not always appear to be called when the component is dismounted.

Here is a minimum viable example:

```python
from uuid import uuid4

from reactpy import component, hooks, html


@component
def child():
    ownership_uuid = hooks.use_memo(lambda: uuid4().hex)

    @hooks.use_effect(dependencies=[])
    async def unmount_manager():
        print("registering new unmount func")
        def unmount():
            # FIXME: This is not working as expected. Dismount is not always called.
            print("unmount func called")
        return unmount

    return ownership_uuid


@component
def root():
    components = hooks.use_ref([child(key=uuid4().hex)])
    reload_trigger, set_reload_trigger = hooks.use_state(True)

    async def add_front_child(event):
        components.current.insert(0, child(key=uuid4().hex))
        set_reload_trigger(not reload_trigger)

    async def add_end_child(event):
        components.current.append(child(key=uuid4().hex))
        set_reload_trigger(not reload_trigger)

    async def remove_front_child(event):
        if components.current:
            components.current.pop(0)
            set_reload_trigger(not reload_trigger)

    async def remove_end_child(event):
        if components.current:
            components.current.pop()
            set_reload_trigger(not reload_trigger)

    return html.div(
        html.div(
            html.button({"onClick": add_front_child}, "Add End"),
            html.button({"onClick": add_end_child}, "Add Front"),
            html.button({"onClick": remove_front_child}, "Remove End"),
            html.button({"onClick": remove_end_child}, "Remove Front"),
        ),
        html.div(f"Components: {components.current}"),
        components.current,
    )

```

You can demo this by adding a bunch of children, then removing them. If you look at terminal, you'll notice that `unmount` was only called once despite multiple children being unloaded.

JamesHutchison:
This might be related to a bug I've seen where nested children are marked as "rendered" and then if there's subsequent updates to their state they get skipped. Changing the rendering order can expose or hide the bug.

If I had to guess, an example of this:
Button alters state on child and triggers callback on parent
Parent alters parameter passed to child based on callback
Child doesn't render with new parameter because it was already rendered.

[...]

PR #1204, opened 2024-03-05
[WIP] Add connection resume, client state tracking, robust reconnect logic, reconnecting hooks, more

JamesHutchison:

## Issues

While in a discussion on the discord, it became apparent that ReactPy was unusable for any application where dynamic scaling is required. For example, in cases where Azure/GCP will start/stop workers for dynamic load balancing. As a result this, users will be disconnected randomly and then would have the state of the ReactPy components reset.

There's also other issues I encountered while building this. Please see my comments (will be added soon) which walk through this very large PR. The intent is not to have this PR merged in all at once, rather, it is a guide to subsequent smaller PRs that can properly add tests, etc. I do not have the time to properly land this PR, so I am asking for help.

## Solution

The solution to the problem was to have the client store the state that's used for rendering components. The theory is that if the client state matches the server state, then things should work just fine.

In reality, there was some challenges to get to that. The first was that IDs of components and events were random instead of deterministic, so this needed to be changed. Now the IDs are based on their patch path.

The next concern is security. While the state isn't obvious to most users, without proper security a client could manipulate variables that an author would have assumed could only be set by the server. An example might be `is_admin` or something, or maybe the user ID of user viewing the page. Another issue is that server secrets might leak to the client if someone isn't careful.

For security, all data given a sha-256 hash as a signature. The server will reject a state variable if it doesn't match the hash. Users are given a personal salt, which they must provide to the server upon reconnection, and the server has a secret "pepper" that is added to make things more difficult. I wasn't satisfied with this, so I added OTP codes that will change, by default, every 4 hours. The OTP codes use site-packages directory's contents (specifically, the parent directory of where reactpy is installed) and the create times to generate a random secret that feeds into the SHA-1 function that generates them. The OTP codes (right now, in the past, and in the future) are added to the hash just to make it that much harder.

All state values have a key that is derived from the file and line number. To avoid leaking this information via brute force, the key is a sha-256 that has about half the bits cut off, making it difficult to recreate.

The client behavior has been revamped. It now will always reconnect, has improved reconnect behavior, and also will disconnect on idle activity (default is 4 minutes). When the user moves the mouse or scrolls, it will reconnect. There's also hooks for when it is reconnecting and has successfully reconnected. The default behavior is to gray out the page and display a spinning pipe character. It's not great.

The performance impact appears to be negligible.

## Checklist

The checklist will be for each individual feature that should be a separate PR. TBD

Work items remaining (this PR):
- [x] Add serialization for datetime types such as timezone and timedelta. This logic is currently in Heavy Resume but makes sense to just give to everyone.
- [x] Fix the server outrunning the client by moving the message handlers out of the tsx file and into the javascript file
- [x] Comment on the PR on everything

## Extracted Issue List
- A - bug fix for when server would return "layout-update" before the client was ready.
   > **Note from _Reactive-Python_**: Can't seem to replicate this bug. Based on logical sequence of operations, this race condition shouldn't exist as long as the receive queue is set up **before** the websocket connection is accepted. Possibly related to ReactPy's websocket consumer on your specific backend.
   > Rather than debugging every backend's websocket consumer, will count this issue as a no-action. This is due to the fact that all websocket code will be rewritten within [PR #1113](https://github.com/reactive-python/reactpy/pull/1113).
- B - bug fix for inconsistent script rendering that resulted in double execution of scripts. Scripts also always are returned to the DOM for user inspection rather than executed directly.
   > **Note from _Reactive-Python_**: See https://github.com/reactive-python/reactpy/pull/1239 
- C - Message updates
   > **Note from _Reactive-Python_**: Not going to merge this since I plan a large overhaul to the client soon.
- D - Have client keep track of state variables (the big one)
   > **Note from _Reactive-Python_**: This is likely better handled via a third party library, rather than shoving this into ReactPy core. We are willing to accept any PRs that add hooks to make this possible.
   > In terms on simply getting `Azure` or `GCP` compatibility, it seems that a "heartbeat" websocket message could be added to our existing design to prevent GCP/Azure from spinning down their workers due to idle timeout.
- E - add ability to debug messages sent and received on the client
   > **Note from _Reactive-Python_**: No changes were found in this PR related to this.
- F - Add idle disconnect time to client
   > **Note from _Reactive-Python_**: Same reasoning as our comment on D
- G - Add connection timeout so that if you connect to a bad node the client doesn't stay stuck to it
   > **Note from _Reactive-Python_**: Same reasoning as our comment on D
- H - Improve message types in client
   > **Note from _Reactive-Python_**: No changes were found in this PR related to this. Will assume it was accidentally folded under issue C.
- I - switch to orjson for faster serialization
   > **Note from _Reactive-Python_**: This PR only utilizes orjson within the sanic backend. But I've made a note to utilize orjson within #1113
- J - Bug fix for corrupted hook state - change hook state to a contextvar
   > **Note from _Reactive-Python_**: See https://github.com/reactive-python/reactpy/pull/1254
- K - Usage of slots
   > **Note from _Reactive-Python_**: There isn't a strong reason to pull out slots at the moment. But I can agree slots usage is a fairly minor optimization.
- L - add priority queue to component rendering - not functioning as expected currently
   > **Note from _Reactive-Python_**: Will ignore these changes since they are noted as broken by PR author.
- M - messages should be a class and not a typed dict
   > **Note from _Reactive-Python_**: No changes were found in this PR related to this.
- N - Use of protocols (substantial performance and usability issues)
   > **Note from _Reactive-Python_**: No changes were found in this PR related to this.
- O - don't use `hasattr` in the manner identified
   > **Note from _Reactive-Python_**: No changes were found in this PR related to this.


Archmonger:
This is definitely way too much within one PR. Can you break this into smaller PRs?

JamesHutchison:
Assuming this headache of mine doesn't turn into something worse, I'll add comments tomorrow. I don't have the capacity to properly break this out, create / update tests, etc for the changes and I'm asking the community for help to land these features. This PR is intended to be a starting point and is not intended to be merged.

Archmonger:
Also just to re-iterate what I said over discord, I've only experienced WS disconnections under high load when ReactPy is not using a [`BACKHAUL_THREAD`](https://reactive-python.github.io/reactpy-django/latest/reference/settings/#reactpy_backhaul_thread).

A potential alternative to client-side state storage is simply migrating that implementation to core.

JamesHutchison:
It's not clear to me how backhaul threads are equivalent to this. Do you have a doc that explains the architecture? My impression from your description was that it helped with stability in django but it wouldn't help if your node count was scaled down and you still had active connections on the terminated node (same with a client's internet disconnecting / reconnecting).

I think having the client manage their own state is a perfect solution. It very much simplifies the infrastructure and reduces costs. There's a slight delay due to the copy of data but since you're reconnecting there's going to be a delay anyways.

src/js/packages/@reactpy/client/src/components.tsx
```diff
@@ -29,7 +29,7 @@ export function Layout(props: { client: ReactPyClient }): JSX.Element {
 
   useEffect(
     () =>
-      props.client.onMessage("layout-update", ({ path, model }) => {
+      props.client.onLayoutUpdate((path: string, model: any) => {
```
JamesHutchison, line 32: Issue A: bug fix for when server would return "layout-update" before the client was ready.

src/js/packages/@reactpy/client/src/components.tsx
```diff
@@ -125,23 +125,15 @@ function ScriptElement({ model }: { model: ReactPyVdom }) {
...
+    const scriptElement: HTMLScriptElement = document.createElement("script");
+    for (const [k, v] of Object.entries(model.attributes || {})) {
+      scriptElement.setAttribute(k, v);
+    }
+    if (scriptContent) {
+      scriptElement.appendChild(document.createTextNode(scriptContent));
     }
-  }, [model.key, ref.current]);
+    ref.current.appendChild(scriptElement);
+  }, [model.key]);
```
JamesHutchison, line 136: Issue B: bug fix for inconsistent script rendering that resulted in double execution of scripts. Scripts also always are returned to the DOM for user inspection rather than executed directly.

The current behavior was to immediately execute a script with no properties. If it had properties, it went to the DOM instead. These different code paths likely hid the bug where scripts were double executing.

The fix for the double execution was to remove `ref.current` from the dependency list

src/js/packages/@reactpy/client/src/messages.ts
```diff
@@ -12,6 +12,11 @@ export type LayoutEventMessage = {
...
 
-export type IncomingMessage = LayoutUpdateMessage;
-export type OutgoingMessage = LayoutEventMessage;
+export type ReconnectingCheckMessage = {
+  type: "reconnecting-check";
+  value: string;
+}
+
+export type IncomingMessage = LayoutUpdateMessage | ReconnectingCheckMessage;
+export type OutgoingMessage = LayoutEventMessage | ReconnectingCheckMessage;
```
JamesHutchison, line 21: Issue C: Message updates. I think looking at it out now I didn't finish this with the most recent messages, such as "state-update"

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -16,6 +16,8 @@ export interface ReactPyClient {
    */
   onMessage(type: string, handler: (message: any) => void): () => void;
 
+  onLayoutUpdate(handler: (path: string, model: any) => void): void;
+
```
JamesHutchison, line 20: Issue A

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -29,16 +31,27 @@ export interface ReactPyClient {
    * @returns A promise that resolves to the module.
    */
   loadModule(moduleName: string): Promise<ReactPyModule>;
+
+  /**
+   * Update state vars from the server for reconnections
+   * @param givenStateVars State vars to store
+   */
+  updateStateVars(givenStateVars: object): void;
```
JamesHutchison, line 39: Issue D: Have client keep track of state variables

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -29,16 +31,27 @@ export interface ReactPyClient {
...
 }
 
 export abstract class BaseReactPyClient implements ReactPyClient {
   private readonly handlers: { [key: string]: ((message: any) => void)[] } = {};
   protected readonly ready: Promise<void>;
   private resolveReady: (value: undefined) => void;
+  protected stateVars: object;
```
JamesHutchison, line 46: Issue D

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -29,16 +31,27 @@ export interface ReactPyClient {
...
+  protected debugMessages: boolean;
```
JamesHutchison, line 47: Issue E: add ability to debug messages sent and received on the client

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -29,16 +31,27 @@ export interface ReactPyClient {
...
+  protected layoutUpdateHandlers: Array<(path: string, model: any) => void> = [];
```
JamesHutchison, line 48: Issue A

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -29,16 +31,27 @@ export interface ReactPyClient {
...
 
   constructor() {
-    this.resolveReady = () => {};
+    this.resolveReady = () => { };
     this.ready = new Promise((resolve) => (this.resolveReady = resolve));
+    this.stateVars = {};
```
JamesHutchison, line 53: Issue D

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -29,16 +31,27 @@ export interface ReactPyClient {
...
+    this.debugMessages = false;
```
JamesHutchison, line 54: Issue E

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -49,9 +62,17 @@ export abstract class BaseReactPyClient implements ReactPyClient {
     };
   }
 
+  onLayoutUpdate(handler: (path: string, model: any) => void): void {
+    this.layoutUpdateHandlers.push(handler);
+  }
+
```
JamesHutchison, line 68: Issue E

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -49,9 +62,17 @@ export abstract class BaseReactPyClient implements ReactPyClient {
...
   abstract sendMessage(message: any): void;
   abstract loadModule(moduleName: string): Promise<ReactPyModule>;
 
+  updateStateVars(givenStateVars: object): void {
+    this.stateVars = Object.assign(this.stateVars, givenStateVars);
+  }
+
```
JamesHutchison, line 75: Issue D

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -65,6 +86,10 @@ export abstract class BaseReactPyClient implements ReactPyClient {
       return;
     }
 
+    if (this.debugMessages) {
+      logger.log("Got message", message);
+    }
+
```
JamesHutchison, line 92: Issue E

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -79,6 +104,9 @@ export abstract class BaseReactPyClient implements ReactPyClient {
 export type SimpleReactPyClientProps = {
   serverLocation?: LocationProps;
   reconnectOptions?: ReconnectProps;
+  idleDisconnectTimeSeconds?: number;
```
JamesHutchison, line 107: Issue F: Add idle disconnect time to client

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -79,6 +104,9 @@ export abstract class BaseReactPyClient implements ReactPyClient {
...
+  connectionTimeout?: number;
```
JamesHutchison, line 108: Issue G: Add connection timeout so that if you connect to a bad node the client doesn't stay stuck to it

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -79,6 +104,9 @@ export abstract class BaseReactPyClient implements ReactPyClient {
...
+  debugMessages?: boolean;
```
JamesHutchison, line 109: Issue E

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -117,14 +145,38 @@ type ReconnectProps = {
   maxRetries?: number;
   backoffRate?: number;
   intervalJitter?: number;
+  reconnectingCallback?: Function;
+  reconnectedCallback?: Function;
```
JamesHutchison: Issue F: Add callbacks with default for when the client is reconnecting

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -117,14 +145,38 @@ type ReconnectProps = {
...
+};
+
+enum messageTypes {
+  isReady = "is-ready",
+  reconnectingCheck = "reconnecting-check",
+  clientState = "client-state",
+  stateUpdate = "state-update",
+  layoutUpdate = "layout-update",
```
JamesHutchison, line 158: Issue H: Improve message types in client

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -136,17 +188,252 @@ export class SimpleReactPyClient
...
+    this.reconnect()
+
+    const reconnectOnUserAction = (ev: any) => {
+      if (!this.isReady && !this.isReconnecting) {
+        this.reconnect();
+      }
+    }
+
+    window.addEventListener('mousemove', reconnectOnUserAction);
+    window.addEventListener('scroll', reconnectOnUserAction);
```
JamesHutchison: Issue D - this reconnects when the user moves their mouse or scrolls the view to try to make the reconnection as invisible as possible

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -136,17 +188,252 @@ export class SimpleReactPyClient
...
+    this.isReconnecting = false;
+    this.isReady = false
+    this.salt = "";
+    this.shouldReconnect = false;
+    this.didReconnectingCallback = false;
+    this.reconnectingCallback = props.reconnectOptions?.reconnectingCallback || this.showReconnectingGrayout;
+    this.reconnectedCallback = props.reconnectOptions?.reconnectedCallback || this.hideReconnectingGrayout;
+
+    this.onMessage(messageTypes.reconnectingCheck, () => { this.indicateReconnect() })
+    this.onMessage(messageTypes.isReady, (msg) => { this.isReady = true; this.salt = msg.salt; });
```
JamesHutchison, line 211: Issue D - the server needs to tell the client when it's done with the handshake process and give the user the salt it is using in hashing. On reconnect the user will provide the server this salt

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -136,17 +188,252 @@ export class SimpleReactPyClient
...
+    this.sleeping = false;
+    this.isReconnecting = false;
+    this.isReady = false
+    this.salt = "";
+    this.shouldReconnect = false;
+    this.didReconnectingCallback = false;
+    this.reconnectingCallback = props.reconnectOptions?.reconnectingCallback || this.showReconnectingGrayout;
+    this.reconnectedCallback = props.reconnectOptions?.reconnectedCallback || this.hideReconnectingGrayout;
+
+    this.onMessage(messageTypes.reconnectingCheck, () => { this.indicateReconnect() })
```
JamesHutchison, line 210: Issue D - the server asks the client if it is reconnecting

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -136,17 +188,252 @@ export class SimpleReactPyClient
...
+    this.onMessage(messageTypes.isReady, (msg) => { this.isReady = true; this.salt = msg.salt; });
+    this.onMessage(messageTypes.clientState, () => { this.sendClientState() });
```
JamesHutchison, line 212: Issue D - the server requests the client state

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -136,17 +188,252 @@ export class SimpleReactPyClient
...
+    this.onMessage(messageTypes.stateUpdate, (msg) => { this.updateClientState(msg.state_vars) });
```
JamesHutchison, line 213: Issue D - dedicated state update message for things that don't cause a layout update. This might not be used anymore. I think I had created this for background events. The reason I don't suggest returning two messages when there's layout updates is because it creates a small chance of the layout and the state vars being desynced.

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -136,17 +188,252 @@ export class SimpleReactPyClient
...
+  }
 
-    this.socket = createReconnectingWebSocket({
-      readyPromise: this.ready,
-      url: this.urls.stream,
-      onMessage: async ({ data }) => this.handleIncoming(JSON.parse(data)),
-      ...props.reconnectOptions,
+  protected invokeLayoutUpdateHandlers(path: string, model: any) {
+    this.layoutUpdateHandlers.forEach(func => {
+      func(path, model);
```
JamesHutchison, line 236: Issue A (fix for layout-update messages outrunning the client)

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -136,17 +188,252 @@ export class SimpleReactPyClient
...
     });
   }
 
-  sendMessage(message: any): void {
-    this.socket.current?.send(JSON.stringify(message));
+  protected showReconnectingGrayout() {
```
JamesHutchison, line 240: Issue F - this is the default callback when the client is reconnecting. It's a gray overlay with a spinning pipe symbol.

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -136,17 +188,252 @@ export class SimpleReactPyClient
...
+    let angle = 0;
+    function spin() {
+      angle = (angle + 2) % 360; // Adjust rotation speed as needed
+      pipeSymbol.style.transform = `rotate(${angle}deg)`;
+      requestAnimationFrame(spin);
+    }
+    spin();
+  }
+
+  hideReconnectingGrayout() {
```
JamesHutchison, line 293: The other default reconnecting callback

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -136,17 +188,252 @@ export class SimpleReactPyClient
...
+    const overlay = document.getElementById('reactpy-reconnect-overlay');
+    if (overlay && overlay.parentNode) {
+      overlay.parentNode.removeChild(overlay);
+    }
+  }
+
+  indicateReconnect(): void {
+    const isReconnecting = this.isReconnecting ? "yes" : "no";
+    this.sendMessage({ "type": messageTypes.reconnectingCheck, "value": isReconnecting }, true)
+  }
```
JamesHutchison: Issue D - this is the logic for when the server requests to know if the client is reconnecting

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -136,17 +188,252 @@ export class SimpleReactPyClient
...
+
+  sendClientState(): void {
+    if (!this.socket)
+      return;
+    this.transmitMessage({
+      "type": "client-state",
+      "value": this.stateVars,
+      "salt": this.salt
+    });
+  }
```
JamesHutchison, line 313: Issue D - this is the client sending over the state and the salt the client was given

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -136,17 +188,252 @@ export class SimpleReactPyClient
...
+
+  updateClientState(stateVars: object): void {
+    if (!this.socket)
+      return;
+    this.updateStateVars(stateVars)
+  }
```
JamesHutchison, line 319: Issue D - core function that updates the state vars on the client

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -136,17 +188,252 @@ export class SimpleReactPyClient
...
+
+  socketLoop(): void {
+    if (!this.socket)
+      return;
+    if (this.messageQueue.length > 0 && this.isReady && this.socket.current && this.socket.current.readyState === WebSocket.OPEN) {
+      const message = this.messageQueue.shift(); // Remove the first message from the queue
+      this.transmitMessage(message);
+    }
+  }
```
JamesHutchison, line 328: Issue D - this allows the client to queue up messages rather than trying to send them immediately. This also acts as a throttle on client messages because this loop is called on an interval

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -136,17 +188,252 @@ export class SimpleReactPyClient
...
+
+  transmitMessage(message: any): void {
+    if (this.socket && this.socket.current) {
+      if (this.debugMessages) {
+        logger.log("Sending message", message);
+      }
```
JamesHutchison, line 334: Issue E - enable logging of messages sent and received on the client

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -136,17 +188,252 @@ export class SimpleReactPyClient
...
+    if (!this.socket)
+      return;
+    if (Date.now() - this.lastMessageTime > this.idleDisconnectTimeMillis) {
+      if (this.socket.current && this.socket.current.readyState === WebSocket.OPEN) {
+        logger.warn("Closing socket connection due to idle activity");
+        this.sleeping = true;
+        this.socket.current.close();
+      }
+    }
+  }
```
JamesHutchison: Issue F - idle timeout check

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -136,17 +188,252 @@ export class SimpleReactPyClient
...
+
+  reconnect(onOpen?: () => void, interval: number = 750, connectionAttemptsRemaining: number = 20, lastAttempt: number = 0): void {
```
JamesHutchison, line 353: Issue D - reconnection with retries is here now

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -136,17 +188,252 @@ export class SimpleReactPyClient
...
+    const intervalJitter = this.reconnectOptions?.intervalJitter || 0.5;
```
JamesHutchison, line 354: The jitter was adjusted because it was capable of taking the interval backwards in time with the prior logic

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -136,17 +188,252 @@ export class SimpleReactPyClient
...
+    const backoffRate = this.reconnectOptions?.backoffRate || 1.2;
+    const maxInterval = this.reconnectOptions?.maxInterval || 20000;
+    const maxRetries = this.reconnectOptions?.maxRetries || 20;
+
+    if (this.layoutUpdateHandlers.length == 0) {
+      setTimeout(() => { this.reconnect(onOpen, interval, connectionAttemptsRemaining, lastAttempt); }, 10);
+      return
+    }
```
JamesHutchison, line 362: Issue A - if there's no layout handlers then don't connect yet.

This should have a comment as it isn't as obvious as it could be of its purpose

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -136,17 +188,252 @@ export class SimpleReactPyClient
...
+      logger.warn("Giving up on reconnecting (hit retry limit)");
+      this.shouldReconnect = false;
+      this.isReconnecting = false;
+      return
+    }
+
+    if (this.shouldReconnect) {
+      // already reconnecting
+      return;
+    }
```
JamesHutchison, line 374: The name `shouldReconnect` could be improved

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -136,17 +188,252 @@ export class SimpleReactPyClient
...
+    lastAttempt = lastAttempt || Date.now();
```
JamesHutchison, line 375: `lastAttempt` is used to reset the retry count and interval. The idea is that if you disconnect too frequently after connecting to a server, then it does not reset the retry / interval. If you get a stable connection for about a minute then it resets.

I think the name and commenting / code structure could be improved around this

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -136,17 +188,252 @@ export class SimpleReactPyClient
...
+          }
+          if (onOpen)
+            onOpen();
+        },
+        onClose: () => {
+          // reset retry interval
+          if (Date.now() - lastAttempt > maxInterval * 2) {
+            interval = 750;
+            connectionAttemptsRemaining = maxRetries;
+          }
```
JamesHutchison: should be refactored into well named a function so what its doing is obvious

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -136,17 +188,252 @@ export class SimpleReactPyClient
...
+      return;
+    }
+    lastAttempt = lastAttempt || Date.now();
+    this.shouldReconnect = true;
+
+    window.setTimeout(() => {
+      if (!this.didReconnectingCallback && this.reconnectingCallback && maxRetries != connectionAttemptsRemaining) {
+        this.didReconnectingCallback = true;
+        this.reconnectingCallback();
+      }
```
JamesHutchison, line 382: Issue F

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -136,17 +188,252 @@ export class SimpleReactPyClient
...
+      this.socket = createWebSocket({
+        connectionTimeout: this.connectionTimeout,
+        readyPromise: this.ready,
+        url: this.urls.stream,
+        onOpen: () => {
+          lastAttempt = Date.now();
+          if (this.reconnectedCallback) {
+            this.reconnectedCallback();
+            this.didReconnectingCallback = false;
+          }
```
JamesHutchison, line 396: Issue F

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -136,17 +188,252 @@ export class SimpleReactPyClient
...
+          if (Date.now() - lastAttempt > maxInterval * 2) {
+            interval = 750;
+            connectionAttemptsRemaining = maxRetries;
+          }
+          lastAttempt = Date.now()
+          this.shouldReconnect = false;
+          this.isReconnecting = true;
+          this.isReady = false;
+          if (this.socketLoopIntervalId)
+            clearInterval(this.socketLoopIntervalId);
```
JamesHutchison, line 412: The socket loop is for sending messages. This stops it

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -136,17 +188,252 @@ export class SimpleReactPyClient
...
+          if (this.idleCheckIntervalId)
+            clearInterval(this.idleCheckIntervalId);
```
JamesHutchison, line 414: Idle timeout interval. This stops it.

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -136,17 +188,252 @@ export class SimpleReactPyClient
...
+          if (!this.sleeping) {
```
JamesHutchison, line 415: Issue F - idle timeout will set it to sleeping

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -136,17 +188,252 @@ export class SimpleReactPyClient
...
+            logger.log(
+              `reconnecting in ${(thisInterval / 1000).toPrecision(4)} seconds... (${newRetriesRemaining} retries remaining)`,
+            );
+            this.reconnect(onOpen, thisInterval, newRetriesRemaining, lastAttempt);
+          }
+        },
+        onMessage: async ({ data }) => { this.lastMessageTime = Date.now(); this.handleIncoming(JSON.parse(data)) },
+        ...this.reconnectOptions,
+      });
+      this.socketLoopIntervalId = window.setInterval(() => { this.socketLoop() }, 30);
```
JamesHutchison: throttle is set to 30 ms for sending messages from the client. I could see making this a configurable parameter

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -136,17 +188,252 @@ export class SimpleReactPyClient
...
+      this.idleCheckIntervalId = window.setInterval(() => { this.idleTimeoutCheck() }, 10000);
```
JamesHutchison, line 428: Idle timeout check is every 10 seconds. No real need to have high accuracy here

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -136,17 +188,252 @@ export class SimpleReactPyClient
...
+
+    }, interval)
+  }
+
+  ensureConnected(): void {
+    if (this.socket.current?.readyState == WebSocket.CLOSED) {
+      this.reconnect();
+    }
+  }
```
JamesHutchison, line 437: This is called to ensure reconnection happens. It's called when you click a button or something that triggers an event

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -136,17 +188,252 @@ export class SimpleReactPyClient
...
+
+  sendMessage(message: any, immediate: boolean = false): void {
+    if (immediate) {
+      this.transmitMessage(message);
+    } else {
+      this.messageQueue.push(message);
+    }
+    this.lastMessageTime = Date.now()
+    this.sleeping = false;
+    this.ensureConnected();
```
JamesHutchison: Issue D - send a message now has a message queue. There's an option to bypass it which is needed for the messages sent prior to the handshake completing. The `isReady` flag lets it know that the server is ready to take event messages.

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -173,25 +460,16 @@ function getServerUrls(props: LocationProps): ServerUrls {
   return { base, modules, assets, stream };
 }
 
-function createReconnectingWebSocket(
+function createWebSocket(
```
JamesHutchison, line 474: Reconnection is owned by the client now

src/js/packages/@reactpy/client/src/reactpy-client.ts
```diff
@@ -247,16 +519,16 @@ function nextInterval(
...
-      backoffRate,
+      backoffRate),
     // don't exceed max interval
     maxInterval,
   );
 }
 
 function addJitter(interval: number, jitter: number): number {
-  return interval + (Math.random() * jitter * interval * 2 - jitter * interval);
+  return interval + (Math.random() * jitter * interval);
```
JamesHutchison, line 542: Remove ability for jitter to take the interval negative

src/py/reactpy/pyproject.toml
```diff
@@ -35,6 +35,9 @@ dependencies = [
   "colorlog >=6",
   "asgiref >=3",
   "lxml >=4",
+  "pyotp",
+  "orjson",
+  "more-itertools",
```
JamesHutchison, line 40: I'm not sure the correct way to add these on this project. I did these by hand.

src/py/reactpy/reactpy/__init__.py
```diff
@@ -1,10 +1,16 @@
...
+    use_location,
+    use_reconnect_effect,
+    use_scope,
+)
 from reactpy.backend.utils import run
 from reactpy.core import hooks
 from reactpy.core.component import component
 from reactpy.core.events import event
 from reactpy.core.hooks import (
+    ReconnectingOnly,
```
JamesHutchison, line 13: Special case for events that should only fire on reconnection

src/py/reactpy/reactpy/__init__.py
```diff
@@ -46,6 +53,7 @@
     "use_effect",
     "use_location",
     "use_memo",
+    "use_reconnect_effect",
```
JamesHutchison, line 56: shorthand function for `use_effect(my_reconnect_func, ReconnectingOnly())`

I don't actually use it yet.

src/py/reactpy/reactpy/backend/sanic.py
```diff
@@ -51,6 +56,7 @@ def configure(
     app: Sanic[Any, Any],
     component: RootComponentConstructor,
     options: Options | None = None,
+    state_recovery_manager: StateRecoveryManager | None = None,
```
JamesHutchison, line 59: Issue D - Heavy Resume uses Sanic so this is the only backend that was updated. The other backends need to be updated

src/py/reactpy/reactpy/backend/sanic.py
```diff
@@ -209,13 +210,13 @@ def _make_send_recv_callbacks(
     socket: WebSocketConnection,
 ) -> tuple[SendCoroutine, RecvCoroutine]:
     async def sock_send(value: Any) -> None:
-        await socket.send(json.dumps(value))
+        await socket.send(orjson.dumps(value).decode("utf-8"))
```
JamesHutchison, line 213: Issue I - switch to orjson for faster serialization

src/py/reactpy/reactpy/core/_life_cycle_hook.py
```diff
@@ -2,12 +2,16 @@
...
+from typing import TYPE_CHECKING, Any, Callable, Protocol, TypeVar
 
 from anyio import Semaphore
 
-from reactpy.core._thread_local import ThreadLocal
 from reactpy.core.types import ComponentType, Context, ContextProviderType
+from reactpy.utils import Ref
+
+if TYPE_CHECKING:
+    from reactpy.core.hooks import _CurrentState
```
JamesHutchison, line 14: Issue D - circular import workaround

src/py/reactpy/reactpy/core/_life_cycle_hook.py
```diff
@@ -18,12 +22,27 @@ async def __call__(self, stop: Event) -> None: ...
 
 logger = logging.getLogger(__name__)
 
-_HOOK_STATE: ThreadLocal[list[LifeCycleHook]] = ThreadLocal(list)
+_hook_state = ContextVar("_hook_state")
```
JamesHutchison, line 25: Issue J - Change hook state to a contextvar

This fixes an issue where the hook stack was being corrupted. I think it was resusing the stack across requests that lived on the same thread.

src/py/reactpy/reactpy/core/_life_cycle_hook.py
```diff
@@ -117,24 +136,46 @@ async def my_effect(stop_event):
...
+        updated_states: dict[str, Any],
+        previous_states: dict[str, Any],
     ) -> None:
         self._context_providers: dict[Context[Any], ContextProviderType[Any]] = {}
         self._schedule_render_callback = schedule_render
         self._scheduled_render = False
         self._rendered_atleast_once = False
         self._current_state_index = 0
-        self._state: tuple[Any, ...] = ()
+        self._state: list = []
```
JamesHutchison, line 160: It's not clear to me why this was a tuple instead of a list. If you're appending, a list is more efficient.

src/py/reactpy/reactpy/core/_life_cycle_hook.py
```diff
@@ -117,24 +136,46 @@ async def my_effect(stop_event):
         "_scheduled_render",
         "_state",
         "component",
+        "reconnecting",
+        "client_state",
+        "_updated_states",
+        "_previous_states",
```
JamesHutchison, line 142: Issue K: I do not think we need to be using slots. Slots provides a minor performance increase and is more for memory when you have a small object. None of the ReactPy logic and objects are trivially small and allocated in large amounts to where this makes a meaningful difference, from what I can tell.

src/py/reactpy/reactpy/core/_life_cycle_hook.py
```diff
@@ -117,24 +136,46 @@ async def my_effect(stop_event):
...
+
+    def add_state_update(self, updated_state: _CurrentState | Ref) -> None:
+        if (
+            updated_state.key
+            and self._previous_states.get(
+                updated_state.key, "__missing_lifecycle_key_value__"
+            )
+            is not updated_state.value
+        ):
+            self._updated_states[updated_state.key] = updated_state.value
```
JamesHutchison, line 178: Issue D - this ensures we're not resending states the client already knows about

src/py/reactpy/reactpy/core/_life_cycle_hook.py
```diff
@@ -157,7 +198,7 @@ def use_state(self, function: Callable[[], T]) -> T:
         if not self._rendered_atleast_once:
             # since we're not initialized yet we're just appending state
             result = function()
-            self._state += (result,)
+            self._state.append(result)
```
JamesHutchison, line 201: follow up for changing _state to a list

src/py/reactpy/reactpy/core/_life_cycle_hook.py
```diff
@@ -232,13 +273,13 @@ def set_current(self) -> None:
         This method is called by a layout before entering the render method
         of this hook's associated component.
         """
-        hook_stack = _HOOK_STATE.get()
+        hook_stack = get_hook_state()
```
JamesHutchison, line 276: Issue J

src/py/reactpy/reactpy/core/_life_cycle_hook.py
```diff
@@ -232,13 +273,13 @@ def set_current(self) -> None:
...
         if hook_stack:
             parent = hook_stack[-1]
             self._context_providers.update(parent._context_providers)
         hook_stack.append(self)
 
     def unset_current(self) -> None:
         """Unset this hook as the active hook in this thread"""
-        if _HOOK_STATE.get().pop() is not self:
+        if get_hook_state().pop() is not self:
```
JamesHutchison, line 284: Issue J

src/py/reactpy/reactpy/core/_thread_local.py
```diff
@@ -1,21 +0,0 @@
-from threading import Thread, current_thread
```
JamesHutchison, line 1: Issue J - this is no longer used

src/py/reactpy/reactpy/core/component.py
```diff
@@ -2,39 +2,61 @@
...
+P = ParamSpec("P")
+
 
 def component(
-    function: Callable[..., ComponentType | VdomDict | str | None]
-) -> Callable[..., Component]:
+    function: Callable[P, T] | None = None,
+    *,
+    priority: int = 0,
+) -> Callable[P, Component]:
```
JamesHutchison: This is from my other PR

https://github.com/reactive-python/reactpy/pull/1199

src/py/reactpy/reactpy/core/component.py
```diff
@@ -2,39 +2,61 @@
...
+T = TypeVar("T", bound=ComponentType | VdomDict | str | None)
+P = ParamSpec("P")
+
 
 def component(
-    function: Callable[..., ComponentType | VdomDict | str | None]
-) -> Callable[..., Component]:
+    function: Callable[P, T] | None = None,
+    *,
+    priority: int = 0,
```
JamesHutchison, line 26: Issue L - add priority queue to component rendering

This seemed unnecessary after implementing it due to how rendering currently works. If the base object renders, then it grabs all the children to go with it. In many cases, components will render at the priority of their parent. I wouldn't work on this unless the way layouts are rendered is redone.

src/py/reactpy/reactpy/core/component.py
```diff
@@ -2,39 +2,61 @@
...
+        if "key" in sig.parameters and sig.parameters["key"].kind in (
+            inspect.Parameter.KEYWORD_ONLY,
+            inspect.Parameter.POSITIONAL_OR_KEYWORD,
+        ):
+            msg = f"Component render function {function} uses reserved parameter 'key'"
+            raise TypeError(msg)
+
+        @wraps(function)
+        def constructor(
+            *args: P.args, key: Any | None = None, **kwargs: P.kwargs
```
JamesHutchison, line 46: typing change is from my pr

src/py/reactpy/reactpy/core/component.py
```diff
@@ -2,39 +2,61 @@
...
 
-    @wraps(function)
-    def constructor(*args: Any, key: Any | None = None, **kwargs: Any) -> Component:
-        return Component(function, key, args, kwargs, sig)
+        return constructor
 
-    return constructor
+    if function:
+        return _component(function)
+    return _component
```
JamesHutchison, line 54: Issue L - this allows for the legacy behavior of `@component` to still work.

`@component`
`@component()`
`@component(priority=-3)`

src/py/reactpy/reactpy/core/component.py
```diff
@@ -2,39 +2,61 @@
...
+    __slots__ = (
+        "__weakref__",
+        "_func",
+        "_args",
+        "_kwargs",
+        "_sig",
+        "key",
+        "type",
+        "priority",
+    )
```
JamesHutchison, line 69: Again, I think Issue K applies here

src/py/reactpy/reactpy/core/component.py
```diff
@@ -43,12 +65,14 @@ def __init__(
...
         kwargs: dict[str, Any],
         sig: inspect.Signature,
+        priority: int = 0,
     ) -> None:
         self.key = key
         self.type = function
         self._args = args
         self._kwargs = kwargs
         self._sig = sig
+        self.priority = priority
```
JamesHutchison, line 85: Issue L

src/py/reactpy/reactpy/core/hooks.py
```diff
@@ -38,6 +43,13 @@
 
 logger = getLogger(__name__)
 
+
+class ReconnectingOnly(list):
+    """
+    Used to indicate that a hook should only be used during reconnection
+    """
```
JamesHutchison, line 50: Issue D

src/py/reactpy/reactpy/core/hooks.py
```diff
@@ -49,7 +61,9 @@ def use_state(initial_value: Callable[[], _Type]) -> State[_Type]: ...
 def use_state(initial_value: _Type) -> State[_Type]: ...
 
 
-def use_state(initial_value: _Type | Callable[[], _Type]) -> State[_Type]:
+def use_state(
+    initial_value: _Type | Callable[[], _Type], *, server_only: bool = False
+) -> State[_Type]:
```
JamesHutchison, line 66: Issue D - `use_state` is client visible by default. `use_ref` is server only by default. Both have a `server_only` argument.

src/py/reactpy/reactpy/core/hooks.py
```diff
@@ -61,23 +75,62 @@ def use_state(initial_value: _Type | Callable[[], _Type]) -> State[_Type]:
...
+def get_caller_info():
+    # Get the current stack frame and then the frame above it
+    caller_frame = sys._getframe(2)
+    for i in range(50):
+        render_frame = sys._getframe(4 + i)
+        patch_path = render_frame.f_locals.get("patch_path_for_state")
+        if patch_path is not None:
+            break
+    # Extract the relevant information: file path and line number and hash it
+    return f"{caller_frame.f_code.co_filename} {caller_frame.f_lineno} {patch_path}"
```
JamesHutchison, line 104: Ugly hack - to preserve the interface, it walks up the stack frames to find `patch_path_for_state` in the locals. I would suggest refactoring it

JamesHutchison: This returns the unique identifier of each state / ref/ hook call based on what called it. This is eventually hashed and then the hash is truncated to create the key. The purpose of truncate the hash is to make it more challenging to reproduce this information.

src/py/reactpy/reactpy/core/hooks.py
```diff
@@ -61,23 +75,62 @@ def use_state(initial_value: _Type | Callable[[], _Type]) -> State[_Type]:
...
+
+
+__DEBUG_CALLER_INFO_TO_STATE_KEY = {}
+
+
+@lru_cache(8192)
+def get_state_key(caller_info: str) -> str:
+    result = hashlib.sha256(caller_info.encode("utf8")).hexdigest()[:20]
+    if __debug__:
+        __DEBUG_CALLER_INFO_TO_STATE_KEY[result] = caller_info
```
JamesHutchison, line 114: When keys are missing, or the value doesn't serialize / deserialize this variable could be used to figure out what its talking about

src/py/reactpy/reactpy/core/hooks.py
```diff
@@ -61,23 +75,62 @@ def use_state(initial_value: _Type | Callable[[], _Type]) -> State[_Type]:
...
+    return result
+
+
 class _CurrentState(Generic[_Type]):
-    __slots__ = "value", "dispatch"
+    __slots__ = "key", "value", "dispatch"
```
JamesHutchison, line 119: Issue K - slots are probably fine here

src/py/reactpy/reactpy/core/hooks.py
```diff
@@ -131,11 +185,17 @@ def use_effect(
...
     last_clean_callback: Ref[_EffectCleanFunc | None] = use_ref(None)
+    hook = get_current_hook()
+    if hook.reconnecting.current:
+        if not isinstance(dependencies, ReconnectingOnly):
+            return
+        dependencies = None
+    else:
+        if isinstance(dependencies, ReconnectingOnly):
+            return
+        dependencies = _try_to_infer_closure_values(function, dependencies)
```
JamesHutchison, line 198: The logic for whether something is executed on reconnection. The `use_memo` and `use_ref` lines must be located where they will always get called

src/py/reactpy/reactpy/core/hooks.py
```diff
@@ -204,7 +264,7 @@ def use_debug_value(
 
     if REACTPY_DEBUG_MODE.current and old.current != new:
         old.current = new
-        logger.debug(f"{current_hook().component} {new}")
+        logger.debug(f"{get_current_hook().component} {new}")
```
JamesHutchison, line 267: Issue J - `current_hook` was renamed to `get_current_hook` to follow naming convention

src/py/reactpy/reactpy/core/layout.py
```diff
@@ -62,34 +76,64 @@ class Layout:
         "_render_tasks_ready",
         "_root_life_cycle_state_id",
         "_model_states_by_life_cycle_state_id",
+        "reconnecting",
+        "client_state",
+        "_state_recovery_serializer",
+        "_state_var_lock",
+        "_hook_state_token",
+        "_previous_states",
```
JamesHutchison, line 87: Issue J - probably shouldn't be using slots

src/py/reactpy/reactpy/core/layout.py
```diff
@@ -62,34 +76,64 @@ class Layout:
...
+        root: ComponentType,
+    ) -> None:
         super().__init__()
-        if not isinstance(root, ComponentType):
-            msg = f"Expected a ComponentType, not {type(root)!r}."
-            raise TypeError(msg)
+        # slow
+        # if not isinstance(root, ComponentType):
+        #     msg = f"Expected a ComponentType, not {type(root)!r}."
+        #     raise TypeError(msg)
```
JamesHutchison, line 101: Issue N - this is very expensive. On my performance trace, nearly half the render time was going to isinstance checks for protocols. I highly recommend removing protocols entirely because they hurt the developer experience since they are disjoint from the actual classes, and because they incur a performance penalty. Just use a base class.

JamesHutchison: I also think these checks should be behind a `if __debug__:` so that we're not wasting cycles in production on something like this.

src/py/reactpy/reactpy/core/layout.py
```diff
@@ -62,34 +76,64 @@ class Layout:
...
         self.root = root
+        self.reconnecting = Ref(False)
```
JamesHutchison, line 103: The ref is significant. When stuff gets created, on reconnection, it just hangs out forever until its reused.

I would prefer refactoring the layout code so that its a bit cleaner and this isn't tied to the state like this. Like you just have a context var that has everything you need to know. However, the current way things are executed get in the way of doing that because stuff just ends up on the main loop in isolation.

src/py/reactpy/reactpy/core/layout.py
```diff
@@ -62,34 +76,64 @@ class Layout:
...
+
+    def set_recovery_serializer(self, serializer: StateRecoverySerializer) -> None:
+        self._state_recovery_serializer = serializer
 
     async def __aenter__(self) -> Layout:
+        return await self.start()
+
+    async def start(self) -> Layout:
+        self._hook_state_token = create_hook_state()
+
```
JamesHutchison, line 116: Issue D - the refactor to `start` is needed so that we can restart the process from scratch if state rebuild fails.

src/py/reactpy/reactpy/core/layout.py
```diff
@@ -122,7 +174,7 @@ async def deliver(self, event: LayoutEventMessage) -> None:
             except Exception:
                 logger.exception(f"Failed to execute event handler {handler}")
         else:
-            logger.info(
+            logger.warning(
```
JamesHutchison, line 180: From what I can tell this shouldn't happen under normal circumstances so I upgraded this to a warning

src/py/reactpy/reactpy/core/layout.py
```diff
@@ -133,6 +185,32 @@ async def render(self) -> LayoutUpdateMessage:
...
+            # and currently there's no (obvious) visibility on if we're waiting for them to finish
+            # so this will wait up to 0.15 * 5 = 750 ms to see if any renders come in before
+            # declaring it done. In the future, it would be better to just track the pending events
+            for _ in range(5):
+                try:
+                    model_state_id = await self._rendering_queue.get_nowait()
+                except asyncio.QueueEmpty:
+                    await asyncio.sleep(0.15)  # make sure
+                else:
+                    break
```
JamesHutchison, line 213: I don't like this but I don't want to put effort into working around this for what is really an exceptional situation. I think this can be properly fixed when layout is refactored to be simpler and more streamlined. You could take hooks that are pending and track them on a contextvar.

src/py/reactpy/reactpy/core/layout.py
```diff
@@ -167,11 +246,21 @@ async def _create_layout_update(
...
+            else {}
+        )
+        self._previous_states.update(updated_states)
+        updated_states.clear()
+        clear_hook_state(token)
+        return LayoutUpdateMessage(
+            type="layout-update",
+            path=new_state.patch_path,
+            model=new_state.model.current,
+            state_vars=state_vars,
```
JamesHutchison, line 265: Issue M - messages should be a class and not a typed dict. Then as a class the type can just be a class variable. I did this refactor so that you can track references using the IDE.

src/py/reactpy/reactpy/core/layout.py
```diff
@@ -188,6 +277,7 @@ async def _render_component(
         await life_cycle_hook.affect_component_will_render(component)
         exit_stack.push_async_callback(life_cycle_hook.affect_layout_did_render)
         try:
+            patch_path_for_state = new_state.patch_path  # type: ignore  # noqa
```
JamesHutchison, line 283: This is the mysterious local variable that's set for the stack frame hack I mentioned above

src/py/reactpy/reactpy/core/layout.py
```diff
@@ -278,7 +368,11 @@ def _render_model_attributes(
             if event in old_state.targets_by_event:
                 target = old_state.targets_by_event[event]
             else:
-                target = uuid4().hex if handler.target is None else handler.target
+                target = (
+                    new_state.patch_path + event
+                    if handler.target is None
+                    else handler.target
+                )
```
JamesHutchison, line 380: Issue D - the target needed to be made deterministic, so its using patch path + event

src/py/reactpy/reactpy/core/layout.py
```diff
@@ -455,14 +565,19 @@ async def _unmount_model_states(self, old_states: list[_ModelState]) -> None:
 
             if model_state.is_component_state:
                 life_cycle_state = model_state.life_cycle_state
-                del self._model_states_by_life_cycle_state_id[life_cycle_state.id]
-                await life_cycle_state.hook.affect_component_will_unmount()
+                try:
+                    del self._model_states_by_life_cycle_state_id[life_cycle_state.id]
+                    await life_cycle_state.hook.affect_component_will_unmount()
+                except KeyError:
+                    pass  # sideeffect of reusing model states
```
JamesHutchison, line 577: Issue D - this `KeyError` check may no longer be needed now that I look at it. Originally I had everything passing around a life cycle state, even non-components, so that you could always pull values from the parent, but now there's a parent state argument for that purpose. As a result, this conditional shouldn't be triggered for non-components since `component_state` is based on whether the life cycle state exists (which is kind of a design issue on its own, but I digress). The side effect of my earlier code was that on clean-up, things would get deleted multiple times here. Again, I don't think it's needed, and adding a breakpoint on that pass should reveal that pretty easily.

src/py/reactpy/reactpy/core/layout.py
```diff
@@ -455,14 +565,19 @@ async def _unmount_model_states(self, old_states: list[_ModelState]) -> None:
...
 
             to_unmount.extend(model_state.children_by_key.values())
 
-    def _schedule_render_task(self, lcs_id: _LifeCycleStateId) -> None:
+    def _schedule_render_task(
+        self, lcs_id: _LifeCycleStateId, priority: int = 0
+    ) -> None:
         if not REACTPY_ASYNC_RENDERING.current:
-            self._rendering_queue.put(lcs_id)
+            self._rendering_queue.put(lcs_id, priority)
```
JamesHutchison, line 585: Issue L - priority queue.

Again, this doesn't work as expected

src/py/reactpy/reactpy/core/layout.py
```diff
@@ -639,13 +788,14 @@ def __init__(
             self._parent_ref = weakref(parent)
             """The parent model state"""
 
-        if life_cycle_state is not None:
-            self.life_cycle_state = life_cycle_state
-            """The state for the element's component (if it has one)"""
+        self.life_cycle_state = life_cycle_state
+        """The state for the element's component (if it has one)"""
```
JamesHutchison, line 797: Issue M - This isn't how you do comments

src/py/reactpy/reactpy/core/layout.py
```diff
@@ -639,13 +788,14 @@ def __init__(
...
+
+        self.parent_life_cycle_state = parent_life_cycle_state
 
     @property
     def is_component_state(self) -> bool:
-        return hasattr(self, "life_cycle_state")
+        return self.life_cycle_state is not None
```
JamesHutchison, line 803: Issue O - don't use `hasattr` like this, it makes for confusing code and IDE's don't pick up this reference

src/py/reactpy/reactpy/core/layout.py
```diff
@@ -663,12 +813,22 @@ def __repr__(self) -> str:  # nocov
...
+    client_state: dict[str, Any],
+    updated_states: dict[str, Any],
+    previous_states: dict[str, Any],
 ) -> _LifeCycleState:
     life_cycle_state_id = _LifeCycleStateId(uuid4().hex)
     return _LifeCycleState(
         life_cycle_state_id,
-        LifeCycleHook(lambda: schedule_render(life_cycle_state_id)),
+        LifeCycleHook(
+            lambda: schedule_render(life_cycle_state_id, component.priority),
```
JamesHutchison, line 831: tbh, this lambda was kind of was a pain in the butt and I recommend refactoring. It's a super important parameter yet you can't find references to where its populated anywhere due to how the code is written. If it's always `schedule_render` then why not just make it part of the class?

src/py/reactpy/reactpy/core/layout.py
```diff
@@ -707,16 +867,21 @@ class _LifeCycleState(NamedTuple):
...
         if value not in self._pending:
             self._pending.add(value)
-            self._loop.call_soon_threadsafe(self._queue.put_nowait, value)
+            self._loop.call_soon_threadsafe(self._queue.put_nowait, (priority, value))
 
     async def get(self) -> _Type:
-        value = await self._queue.get()
+        priority, value = await self._queue.get()
+        self._pending.remove(value)
+        return value
```
JamesHutchison, line 886: Issue L - priority queue, which again, didn't work as expected

src/py/reactpy/reactpy/core/layout.py
```diff
@@ -707,16 +867,21 @@ class _LifeCycleState(NamedTuple):
...
+
+    async def get_nowait(self) -> _Type:
+        priority, value = self._queue.get_nowait()
         self._pending.remove(value)
         return value
```
JamesHutchison, line 891: Issue D - `get_nowait` was needed because the reconnection logic needed to not block.

src/py/reactpy/reactpy/core/layout.py
```diff
@@ -729,7 +894,7 @@ def _get_children_info(children: list[VdomChild]) -> Sequence[_ChildInfo]:
         elif isinstance(child, dict):
             child_type = _DICT_TYPE
             key = child.get("key")
-        elif isinstance(child, ComponentType):
+        elif isinstance(child, (Component, _ContextProvider)):
```
JamesHutchison, line 902: Issue N - performance refactor

src/py/reactpy/reactpy/core/serve.py
```diff
@@ -8,16 +10,39 @@
...
 
 
-SendCoroutine = Callable[[LayoutUpdateMessage], Awaitable[None]]
+SendCoroutine = Callable[
+    [
+        LayoutUpdateMessage
+        | ReconnectingCheckMessage
+        | IsReadyMessage
+        | ClientStateMessage
+    ],
```
JamesHutchison, line 38: Not a fan of this typing, should just be refactored to use a base type

src/py/reactpy/reactpy/core/serve.py
```diff
@@ -40,44 +65,146 @@ async def serve_layout(
...
+                raise
+        finally:
+            clear_hook_state(token)
 
 
 async def _single_incoming_loop(
     task_group: TaskGroup,
     layout: LayoutType[LayoutUpdateMessage, LayoutEventMessage],
     recv: RecvCoroutine,
+    send: SendCoroutine,
```
JamesHutchison, line 108: This is taking in `send` but its not used, so we can remove this. Originally the idea was that when events occurred it would send off state updates separate from layout updates, however it turned out that the isolation was too high. Again, a refactor on layout would be preferable.

src/py/reactpy/reactpy/core/serve.py
```diff
@@ -40,44 +65,146 @@ async def serve_layout(
...
 ) -> None:
     while True:
         # We need to fire and forget here so that we avoid waiting on the completion
         # of this event handler before receiving and running the next one.
         task_group.start_soon(layout.deliver, await recv())
+
+
+class WebsocketServer:
```
JamesHutchison, line 116: This is all Issue D

On initial connection, the client establishes that it is not reconnecting. The server generates a salt and hands it to the client in the `is-ready` message. The server then renders the layout

On reconnection, the client establishes it is reconnecting and provides the previous salt as well as its client state. The server does a reconnection render to rebuild the state, using the client state to populate initial values. This can trigger async events which take time to finish to fully rebuild the state. After the rebuild finishes, the server sends the `is-ready` message.

If a hash doesn't match up then a `StateRecoveryFailureError` will be raised

src/py/reactpy/reactpy/core/state_recovery.py
```diff
@@ -0,0 +1,295 @@
+import asyncio
```
JamesHutchison, line 1: This file is all Issue D

src/py/reactpy/reactpy/core/state_recovery.py
```diff
@@ -0,0 +1,295 @@
...
+class StateRecoveryFailureError(Exception):
+    """
+    Raised when state recovery fails.
+    """
+
+
+class StateRecoveryManager:
+    def __init__(
+        self,
+        serializable_types: Iterable[type],
```
JamesHutchison, line 31: the list of allowed types. Each _root_ type must be provided here. If you have nested values that are never a root type, you don't need to bother. For example, I have a pydantic class that's 5 layers nested across 12 types, but I only need to provide the top level one.

src/py/reactpy/reactpy/core/state_recovery.py
```diff
@@ -0,0 +1,295 @@
...
+        pepper: str,
```
JamesHutchison, line 32: This is a secret added to the hash, provided by the server. It can be retrieved from any source.

src/py/reactpy/reactpy/core/state_recovery.py
```diff
@@ -0,0 +1,295 @@
...
+        otp_key: str | None = None,
```
JamesHutchison, line 33: An OTP key. If one is not provided, it is generated by listing out the directories under which reactpy is installed and grabbing the creation time. The OTP values are added to the hash

src/py/reactpy/reactpy/core/state_recovery.py
```diff
@@ -0,0 +1,295 @@
...
+        otp_interval: int = (4 * 60 * 60),
```
JamesHutchison, line 34: The interval in seconds for which the OTP will rotate. This is 4 hours

src/py/reactpy/reactpy/core/state_recovery.py
```diff
@@ -0,0 +1,295 @@
...
+        otp_digits: int = 10,  # 10 is the max allowed
+        otp_max_age: int = (48 * 60 * 60),
```
JamesHutchison, line 36: When a client fails to provide a matching signature, different OTP times are tried. This is the maximum age to try, in seconds. This defaults to 48 hours

src/py/reactpy/reactpy/core/state_recovery.py
```diff
@@ -0,0 +1,295 @@
...
+        # OTP code is actually three codes, in the past and future concatenated
+        otp_mixer: float = (365 * 24 * 60 * 60 * 3),
```
JamesHutchison, line 38: To add more entropy, three OTP codes are generated. Right now, in the past, and in the future. This is defining the distance to grab the past and future, which defaults to 3 years

src/py/reactpy/reactpy/core/state_recovery.py
```diff
@@ -0,0 +1,295 @@
...
+        max_num_state_objects: int = 512,
```
JamesHutchison, line 39: The maximum number of objects to send the client in a single state update

src/py/reactpy/reactpy/core/state_recovery.py
```diff
@@ -0,0 +1,295 @@
...
+        max_object_length: int = 40000,
```
JamesHutchison, line 40: The maximum length of a serialized object to send to the client

src/py/reactpy/reactpy/core/state_recovery.py
```diff
@@ -0,0 +1,295 @@
...
+        default_serializer: Callable[[Any], bytes] | None = None,
```
JamesHutchison, line 41: This is slightly misleading. When serializing and an object not supported by `orjson` is encountered, it falls back to a default. However, a default is already provided for types like timezone, timedelta, and dataclasses. If the provided default cannot serialize the object, then this is the fallback for that.

src/py/reactpy/reactpy/core/state_recovery.py
```diff
@@ -0,0 +1,295 @@
...
+        deserializer_map: dict[type, Callable[[Any], Any]] | None = None,
```
JamesHutchison, line 42: A map of object to deserialization. By default, `TheObjectType(**kwargs)` is tried which probably won't work for nested objects, but might.

src/py/reactpy/reactpy/core/state_recovery.py
```diff
@@ -0,0 +1,295 @@
...
+
+    def _serialize(self, key: str, obj: object) -> tuple[str, str, str]:
+        type_id = b"1"  # bool
+        if obj is None:
+            return "0", "", ""
+        match obj:
+            case True:
+                result = b"true"
+            case False:
+                result = b"false"
```
JamesHutchison, line 175: This extra complexity is for performance reasons.

`None` is not signed.
Boolean values are really common so this is a shortcut.

src/py/reactpy/reactpy/core/vdom.py
```diff
@@ -328,7 +328,7 @@ def _validate_child_key_integrity(value: Any) -> None:
         )
     else:
         for child in value:
-            if isinstance(child, ComponentType) and child.key is None:
+            if child.key is None and isinstance(child, ComponentType):
```
JamesHutchison, line 331: Issue N - perf: since this is a warning condition, the cheaper check is done first.

src/py/reactpy/reactpy/utils.py
```diff
@@ -27,12 +27,26 @@ class Ref(Generic[_RefValue]):
...
+        self.key = key
+        self._hook = None
+        if key:
+            hook = get_current_hook()
+            hook.add_state_update(self)
+            self._hook = hook
+
+    @property
+    def value(self) -> _RefValue:
+        return self.current
```
JamesHutchison, line 49: aligns the interface with `use_state` for serialization

src/py/reactpy/reactpy/utils.py
```diff
@@ -27,12 +27,26 @@ class Ref(Generic[_RefValue]):
         You can compare the contents for two ``Ref`` objects using the ``==`` operator.
     """
 
-    __slots__ = ("current",)
+    __slots__ = ("current", "key", "_hook")
+
+    def __init__(
+        self, initial_value: _RefValue = _UNDEFINED, key: str | None = None
+    ) -> None:
+        from reactpy.core._life_cycle_hook import get_current_hook
```
JamesHutchison, line 35: IIRC this works around a circular import

JamesHutchison:
My suggestion on approach:
- Keep the javascript client updates and fixes as a single PR, even though it resolves multiple issues. It either needs to be backwards compatible or could merge into the PR that does the server side of Issue D
- Some of these are small but big wins. I would do them first. For example, fixing the performance problem caused by the protocols `isinstance` checks is only a couple lines of code
- Some are worth discussing before action. The priority queue sounds good but doesn't actually work as expected. The render times are also really small so the value of a priority queue is questionable. It might be useful in the future though.
- Measure using a tool like py-spy or something to get performance bottlenecks rather than making any assumptions. I thought the sha256 calls would show up somewhere but they, and their parent functions, haven't.

JamesHutchison:
Example usage that supports Pydantic / SQLModel objects:

```python
def default_serializer(obj: Any) -> Any:
    if isinstance(obj, BaseModel):
        return obj.dict()
    raise TypeError(f"Object of type {obj.__class__.__name__} is not JSON serializable")


state_recovery_manager = StateRecoveryManager(
    serializable_types=[
        TemporaryUser,
        QAMessage,
    ],
    pepper="pepper",
    default_serializer=default_serializer,
)

configure(app, Main, Options(head=head, cors=cors_options), state_recovery_manager)
```

JamesHutchison:
If you want to see this in action, it is live on heavyresume.com

Inactivity timeout is set to 4 minutes. You can see it happen in the console. If you scroll down and let it time out, you'll notice that it doesn't regenerate the view on reconnect. Moving the mouse or scrolling triggers reconnection.

Thanks!

JamesHutchison:
One thing I'd like to bring attention to - if the layout logic is rewritten, please leave some means to re-render all components of a certain type. Next best thing is some means to trigger a whole page re-render from the client state. The use case would be hot reloading. Right now I can successfully use jurigged to hot reload - however unless the component has a toggle or something to force a re-render, I have to reload the page.

edit: this is no longer true with my PR below. That PR will disconnect and reconnect and force a rerender.

JamesHutchison:
This PR (on my fork) leverages the logic from this PR to add hot reloading. I will not be merging it into my main branch because I don't want to clutter things up further, and I'm fine just checking out that one branch in the mean time.

https://github.com/JamesHutchison/reactpy/pull/12

JamesHutchison:
This linked PR fixes a bug where `use_effect(my_func, [])` was rerunning after the next reconnection. The fix from that PR applies to this one.

https://github.com/JamesHutchison/reactpy/pull/16

src/js/packages/@reactpy/client/src/components.tsx
Archmonger: Does this issue only exist due to the client state management stuff in this PR?

src/py/reactpy/reactpy/core/_life_cycle_hook.py
Archmonger: Does this issue only exist due to the client state management stuff in this PR?

src/js/packages/@reactpy/client/src/components.tsx
JamesHutchison: It's been a long while. My intuition is "no" but one would need to check the code. IIRC there's a delay from connecting to setting up the handlers and no way to set up the handlers ahead of time, but this is top of my head and I don't have the code in front of me.

src/py/reactpy/reactpy/core/_life_cycle_hook.py
JamesHutchison: No, I think the issue was that Sanic reuses the same thread for multiple requests. ContextVars were designed for this very problem.

JamesHutchison:
After using the heavy stack for many months, here's the known issues that are currently unresolved and may be attributed to my changes:

- pokey loading on slow connections - seems to be caused by layout updates requiring confirmation from the client before moving forward.
- imperfect connection recovery - connection recovery for larger states always fail to rebuild in time resulting in destructive reconstruction that can result in state loss for text boxes that have unsaved changes
- input hell - inputs and components that have matching DOM paths aren't recreated unless the key changes or the DOM structure leading up to them changes. This means, for example, if you change a selected list item, the inputs won't update even though the parameters into the component do
- incomplete component updates - in some cases the render order of components results in child components not rendering after the parent finishes. For example, a parent triggers a child render, the parent is then rendered, perhaps due to an event or state change, such that the child needs a rerender, but the child is marked as having already rendered and does not render again.

I'm pretty sure the input hell stems from the change that gives state a deterministic target based on the DOM path. This change was needed to support state rebuild.

The rest may be existing ReactPy issues.

src/js/packages/@reactpy/client/src/components.tsx
Archmonger: I'm testing out this implementation in https://github.com/reactive-python/reactpy/pull/1239 but it seems that it breaks script re-execution when the script `key` (or `content`) changes.

JamesHutchison: That sounds like the issue I've seen with inputs. You'll change the key and value on the input but existing value sticks around. It's an issue if you have a list of objects with inputs, update the list, then redraw. The browser shows the old value and doesn't update it.

The workaround IIRC is to wrap it and change the key on what its wrapped in.

I was under the impression my code change to support reconnection was the cause but maybe its not? If that's the case, it sounds like its a bug that affects everything, not just scripts.

JamesHutchison: Also, just to call this out, I'm not sure having script tags execute multiple times because they changed is necessarily a good practice. I can think of a couple scenarios where implementation changes such as how renders are sent to the client might create behavioral changes due to "in-between" states getting skipped. Seems like it should always be one script one execution.

Archmonger: I could agree with that - Given that sequential renders that affect the same components are "optimized out" by `layout.py`, there is definitely a reality where things could get skipped.

I am also aware that `layout.py` has some bugs, so it's on my plate to rewrite that.

See issue:
- https://github.com/reactive-python/reactpy/issues/1198

Archmonger:
- pokey loading on slow connections
   - Likely due to your modifications 
- imperfect connection recovery
   - Possibly related to the fact that the server currently sends messages as a single request, and does not attempt to stream large requests into chunks? Would need more debugging, but I haven't seen this issue yet.
- input hell
   -  Agreed this is likely due to your modifications
- incomplete component updates
   - I haven't seen this issue in any of my stress tests, but `layout.py` is in rough shape so it's possible some of your changes exposed a logic deficiency (probably unsafe `layout.py` assumptions).

Archmonger:
New PRs have been created where necessary, and co-author credit was provided on these new PRs.

Going to close this.

PR #1210, opened 2024-03-07
Move `reactpy.backend.hooks` module into `reactpy.core.hooks`

### LINKED ISSUE #1182: Remove `reactpy.backend.hooks`
### Current Situation

It is currently a bit disjointed that some of our hooks exist within the `reactpy.backend.hooks` module.

### Proposed Actions

Move all hooks within `reactpy.backend.hooks` into `reactpy.hooks`.

joshbmair:
## Description

Fixes #1182 - Moves the hooks in `reactpy.backend.hooks` into `reactpy.core.hooks`

## Checklist

-   [X] Tests have been developed for bug fixes or new functionality.
-   [X] The changelog has been updated, if necessary.
-   [X] Documentation has been updated, if necessary.
-   [X] GitHub Issues closed by this PR have been linked.


src/py/reactpy/reactpy/core/hooks.py
```diff
@@ -248,6 +263,16 @@ def use_context(context: Context[_Type]) -> _Type:
...
+def use_scope() -> MutableMapping[str, Any]:
+    """Get the current :class:`~reactpy.backend.types.Connection`'s scope."""
+    return use_connection().scope
+
+
+def use_location() -> Location:
+    """Get the current :class:`~reactpy.backend.types.Connection`'s location."""
+    return use_connection().location
+
+
```
Archmonger, line 274: Minor nitpick, but could we move these hooks all be next to each other?

Right now, `use_connection` isn't sitting next to the rest of these hooks.

Archmonger (changes requested):
Can be merged after resolving my minor nitpick.

src/py/reactpy/reactpy/backend/hooks.py
```diff
@@ -1,30 +0,0 @@
-from __future__ import annotations
```
Archmonger, line 1: Oh - Actually one more thing. I'd like to release this PR within the next minor rev (rather than immediately creating a major release).

To do that, we are going to need to keep a copy of the functions in `reactpy.backends.hooks` and provide a [deprecation warning](https://github.com/reactive-python/reactpy/blob/4307a09dfa75e6c1b10a285e5ae4bdf0323cd018/src/py/reactpy/reactpy/core/vdom.py#L177-L181).

This deprecation also needs to be added in the changelog.

Archmonger (changes requested):
Needs some formatting fixes to pass linting tests.

Also, coverage is failing right now. You'll need to convert the existing tests to use your new hooks, and add `# nocov` to the top of the deprecated file.

joshbmair:
I'm having trouble with adding the # nocov comment at the top of backend.hooks. It seems to only be working on the first import line instead of the whole file.

Archmonger:
Hmm I thought there was a way to do it for the whole file. Worst case you could add that comment to each individual hook in the file.

joshbmair:
I decided to add nocov comments to every import/function. Let me know if there's a more concise way to do it, because I couldn't figure it out.

src/py/reactpy/reactpy/backend/hooks.py
```diff
@@ -1,30 +1,45 @@
...
+
     return use_connection().scope
 
 
-def use_location() -> Location:
+def use_location() -> Location:  # nocov
     """Get the current :class:`~reactpy.backend.types.Connection`'s location."""
+    warn(
+        "The module reactpy.backend.hooks has been deprecated and will be deleted in the future. ",
+        "Call use_location in reactpy.core.hooks instead.",
```
Archmonger: Since the hook is exposed at our top-level module, we may as well say: `Call reactpy.use_location instead.`

Same deal with the other functions.

Archmonger:
Still not passing lint:
`reactpy/core/hooks.py:504: error: Incompatible return value type (got "Sequence[Any] | EllipsisType | None", expected "Sequence[Any] | None")  [return-value]**`**

EDIT: Might be unrelated to your changes? If so, you can pin `mypy` to the previous version that doesn't break tests.

joshbmair:
Why do you mean by pin mypy?

joshbmair:
I don't see how my changes could have caused the lint error we're seeing, so I think I'm good to add `mypy` to the most recent version that doesn't break tests (once I know how to do that).

PR #1224, opened 2024-07-18
Fixes needless custom component recreation (#1195)

### LINKED ISSUE #1195: Client-side component events break input focus
### Discussed in https://github.com/reactive-python/reactpy/discussions/1187

<sup>Originally posted by **williamneto** January 13, 2024</sup>
So, while developing [reactpy-material](https://github.com/williamneto/reactpy-material) i faced a problem that might be related to the reactpy architecture for custom components. 

To showcase the problem i created [this](https://github.com/williamneto/reactpy-events-problem) small component, just a normal textfield. The problem is when attaching a on_change event to this custom component, the event on the page doesn't behavior as expected. Each time you type in the field (and trigger the onChange event) the page focus is moved out of the input element.

So, a simple custom TextField like that
```javascript
import React from "react";
import ReactDOM from "react-dom";
import htm from "htm";

const html = htm.bind(React.createElement);

export function bind(node, config) {
  return {
    create: (type, props, children) => React.createElement(type, props, ...children),
    render: (element) => ReactDOM.render(element, node),
    unmount: () => ReactDOM.unmountComponentAtNode(node),
  }
}

export function TextField(attrs) {
  return html`
    <div>
      <input type="text" ...${attrs}>
    </div>`;
}
```

```python
from pathlib import Path
from typing import Any
from reactpy.web.module import export, module_from_file
from reactpy import component

_js_module = module_from_file(
    "reactpy-events-problem",
    file=Path(__file__).parents[0] / "bundle.js"
)

new_text_field = export(_js_module, "TextField")

@component
def text_field(attrs: Any ={}):
    return new_text_field(attrs)
```

And when you try to use it like that

```python
from reactpy import component, run, html, use_state
from reactpy_events_problem import text_field

@component
def app():
    text, set_text = use_state("Text")

    return html.div(
        html.h1(text),
        text_field(
            attrs={
                "value": text,
                "onChange": lambda e: set_text(e["target"]["value"])
            }
        )
    )

run(app)
```

You end up with this weird behavior at the page focus, making type in this field very unpratical</div>

shawncrawley:
As detailed in issue #1195, custom components were being needlessly re-created on every re-render. This was due to the `vdomImportSource` (i.e. `model.importSource`) of the custom component being used as a dependency of the `useEffect` function that creates/binds the custom components. Using a JavaScript object (i.e. dictionary) as a dependency was an issue in this case since the object in question was being recreated anew every re-render even though its actual key/value entries were not. 

I initially completely removed `vdomImportSource` from the dependencies, not understanding a use case where it would ever change anyway, but second guessed its importance due to it being there in the first place. I ended up finding [this article](https://profy.dev/article/react-useeffect-with-object-dependency) and opted for their recommended "Approach 4", which is to `JSON.stringify` the object and use that as the dependency instead.

I also went ahead and refactored the `react.js` web template file to use the [new createRoot(...).render() workflow in ReactJS 18](https://react.dev/reference/react-dom/client/createRoot#reference), as mentioned by @rmorshea in the original issue. Although it seems to be working great for me, I'm definitely not a ReactPy nor even a ReactJS expert, so that should definitely be looked over by someone more experienced.

## Checklist

### DIFF
--- a/docs/source/about/changelog.rst
+++ b/docs/source/about/changelog.rst
@@ -23,6 +23,7 @@ Unreleased
 - :pull:`1118` - `module_from_template` is broken with a recent release of `requests`
 - :pull:`1131` - `module_from_template` did not work when using Flask backend
 - :pull:`1200` - Fixed `UnicodeDecodeError` when using `reactpy.web.export`
+- :pull:`1224` - Fixes needless unmounting of JavaScript components during each ReactPy render.
 
 **Added**
 
--- a/src/js/packages/@reactpy/client/src/components.tsx
+++ b/src/js/packages/@reactpy/client/src/components.tsx
@@ -177,7 +177,7 @@ function useForceUpdate() {
 
 function useImportSource(model: ReactPyVdom): MutableRefObject<any> {
   const vdomImportSource = model.importSource;
-
+  const vdomImportSourceJsonString = JSON.stringify(vdomImportSource);
   const mountPoint = useRef<HTMLElement>(null);
   const client = React.useContext(ClientContext);
   const [binding, setBinding] = useState<ImportSourceBinding | null>(null);
@@ -203,7 +203,7 @@ function useImportSource(model: ReactPyVdom): MutableRefObject<any> {
         binding.unmount();
       }
     };
-  }, [client, vdomImportSource, setBinding, mountPoint.current]);
+  }, [client, vdomImportSourceJsonString, setBinding, mountPoint.current]);
 
   // this effect must run every time in case the model has changed
   useEffect(() => {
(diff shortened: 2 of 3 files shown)


shawncrawley:
I'm not sure on those checklist items above,. It seemed at least a few wouldn't be relevant here, but let me know if I'm wrong there and I'm happy to wrap them up.

Archmonger:
Tests are failing, due to a mixture of unpinned test dependencies and using old/deprecated github workflows for NodeJS setup.

You are free to do some quick fixes to get the tests functioning, or wait for me to fix them in a different PR. Hopefully I'll have time in the next week or so.

Archmonger:
cc: @williamneto

PR #1271, opened 2025-02-10
Ensures key is properly propagated to importedElements

### LINKED ISSUE #1275: The key attribute is not being properly propagated and applied to importedElements
### Discussed in https://github.com/reactive-python/reactpy/discussions/1270

<sup>Originally posted by **shawncrawley** February 10, 2025</sup>
I could not get [GridLayout](https://github.com/react-grid-layout/react-grid-layout?tab=readme-ov-file) to work as a [dynamically loaded component](https://reactpy.dev/docs/guides/escape-hatches/javascript-components.html#dynamically-loaded-components). After digging into the issue, I found that it was because the children of a GridLayout **_must_** have the key attribute, but it is not being found on the children when constructing the GridLayout object on the client side. I am clearly passing it in on the Python side, as seen in my example below, so it is getting dropped/lost somewhere.

```python
return GridLayout(
    Props(
        className="layout",
        layout=layout,
        cols=12,
        rowHeight=30,
        width=1200,
    ),
    lib.html.div(Props(key="a"), "a"),
    lib.html.div(Props(key="b"), "b"),
    lib.html.div(Props(key="c"), "c")
)
```

shawncrawley:
Fixes #1270. The "key" prop was being lost in the trickle-down of importedElements.

## Description

I just added a block of code to the createAttributes function to check if the "key" is part of the vdom model and if so, to add it to the attributes. There may be a better and more thorough way to do this higher up the call chain, but this fixed the issue for me in my use case (see #1270).

## Checklist


Archmonger:
Can you add a test for this within `src/reactpy/tests/test_web/test_module.py`?

You can directly use the `GridLayout` component if you'd like.

shawncrawley:
@Archmonger I'm fresh off of vacation and will try to get that test incorporated ASAP. Thanks for already taking a look at this.

shawncrawley:
So I never actually set up a proper development environment for reactpy. Thus, I've never ran tests or anything. I just barely tried to follow the [Contribute Guide](https://reactpy.dev/docs/about/contributor-guide.html) to get the test environment set up, and am running into issues.

I'm on a Windows 11 machine and used conda to create a new virtual environment with Python 3.13.2. I installed all of the prerequisites. Then when running `hatch test` I get the following error:

```
ImportError while loading conftest 'C:\Users\shawn\Code\reactpy\tests\conftest.py'.
tests\conftest.py:11: in <module>
    from reactpy.config import (
E   ModuleNotFoundError: No module named 'reactpy'
```

I tried to install reactpy from the cloned repo using `pip install -e .` and although that seems to work without failing, I continue to get the `ModuleNotFoundError` above...

Any help would be appreciated. Thanks!

Archmonger:
Have you tried using standard CPython venv instead of using conda (ex `python -m venv .venv`).

I haven't tried our toolchain with Conda yet, so there might be some deficiencies there.

shawncrawley:
Got it working! And I have the test written. My first stab at writing a test within this framework. Unlike the other examples in `test_module.py` I had to actually use `react` and `react-dom` instead of `preact` for it to work as expected. With `preact` I would get an error in the test browser console - something about objects not being able to be passed as children to GridLayout (i.e. the reactpy.html.div children elements were not being properly converted to the expected objects). Let me know if you see any issues with this.

tests/test_web/test_module.py
```diff
@@ -208,6 +208,57 @@ async def test_imported_components_can_render_children(display: DisplayFixture):
         assert (await child.get_attribute("id")) == f"child-{index + 1}"
 
 
+async def test_keys_properly_propagated(display: DisplayFixture):
```
Archmonger, line 211: Can you add a docstring to this that references the issue link?

docs/source/about/changelog.rst
```diff
@@ -69,6 +70,7 @@ Unreleased
 **Fixed**
 
 - :pull:`1239` - Fixed a bug where script elements would not render to the DOM as plain text.
+- :pull:`1271` - Fixed bug where JavaScript components were unable to obtain the ``key`` attribute provided within ReactPy.
```
Archmonger: Can you double check that my changelog entry of this issue is characterized correctly?

Archmonger (changes requested):

Archmonger:
I'm okay with using React for that test. I've seen oddball scenarios where mixed preact/react contexts are hit or miss. That might be the cause of issues with using a parent `preact` component with `react-grid-layout` children.

Out of curiosity, was the issue with setting up your development environment related to Conda?

docs/source/about/changelog.rst
shawncrawley: I tweaked the verbiage a bit, should be good to go.

tests/test_web/test_module.py
shawncrawley: Done.

tests/test_web/test_module.py
```diff
@@ -208,6 +208,64 @@ async def test_imported_components_can_render_children(display: DisplayFixture):
...
+    """ Addresses issue #1271
+    The "key" property was being lost in its propagation from
+    the server-side ReactPy definition to the front-end JavaScript.
+    This is a critical issue for the proper functionality of
+    certain components, such as the GridLayout from 
+    react-grid-layout.
+    """
```
Archmonger:
```suggestion
    """
    Fix https://github.com/reactive-python/reactpy/issues/1275

    The `key` property was being lost in its propagation from the server-side ReactPy
    definition to the front-end JavaScript.
    
    This property is required for certain JS components, such as the GridLayout from 
    react-grid-layout.
    """
```

shawncrawley:
> I'm okay with using React for that test. I've seen oddball scenarios where mixed preact/react contexts are hit or miss. That might be the cause of issues with using a parent `preact` component with `react-grid-layout` children.
> 
> Out of curiosity, was the issue with setting up your development environment related to Conda?

I'm quite sure it was not related to `conda`. There were two indisputable issues, however. The first was a permission denied error that occurred when the underlying `hatch run build:javascript` command was running and attempting to cache some packages. I had to run my Command Prompt as Administrator to get past that.

The second issue was much more difficult to track down. I got that same permission error above whether I used `conda` or the system Python. In both cases, I immediately re-opened the Command Prompt as Administrator and re-executed the `hatch test` command. However, this time it would no longer fail on the permission denied error, but would instead throw the `reactpy ModuleNotFoundError`. Which is where I was at when I messaged you about it above. 

I eventually determined - my best guess at least - that the underlying isolated test environment must have been getting cached and treated as built even though it had failed with that first permissions error, if that makes sense. All I know is that to finally get it to work, I had to execute the `hatch env prune` command, after which the `hatch test` command (as Admin) finally worked. 

So again, apparently if something in the underlying build fails when running `hatch test` for the first time, a manual `hatch env prune` must be done to recover before rebuilding after fixing the issue with the failed build. That seems like a bug with `hatch` to me... In fact, after typing that last sentence, I went and searched for it and found [this open issue](https://github.com/pypa/hatch/issues/908) that seems to match what I've described - and I commented on it as well. So there you have it. Way bigger pain than it should have been for sure.

Archmonger:
Hatch will always caches environments it creates, which is expected. But Hatch really should have auto-deleted the initial environment if the build command failed.

I agree that issue you linked seems to be the culprit. Hopefully they fix it soon.

Archmonger (commented):
PR looks good and likely can be merged as is.

However, before merging can you see if removing the `pop` call from [this LOC](https://github.com/reactive-python/reactpy/blob/fa48bbcd49610fefa9295abadeab625a1d0a2fb7/src/reactpy/core/vdom.py#L151) is a better solution to this issue?

src/reactpy/core/vdom.py
```diff
@@ -148,7 +148,7 @@ def __call__(
     ) -> VdomDict:
         """The entry point for the VDOM API, for example reactpy.html(<WE_ARE_HERE>)."""
         attributes, children = separate_attributes_and_children(attributes_and_children)
-        key = attributes.pop("key", None)
+        key = attributes.get("key", None)
```
shawncrawley, line 151: @Archmonger Per your suggestion, how does this look? Test passed. Seems like it should maintain all of the same behavior and is a cleaner solution.

shawncrawley: Oh, but other tests do fail... Should've been thorough with that. Let me look into what failed and why.

shawncrawley:
There we go. I have the server-side implementation complete with tests updated and passing. Let me know if you prefer the original client-side way I had done it, and I can go ahead and revert.

src/reactpy/_html.py
```diff
@@ -104,7 +104,7 @@ def _fragment(
     event_handlers: EventHandlerDict,
 ) -> VdomDict:
     """An HTML fragment - this element will not appear in the DOM"""
-    if attributes or event_handlers:
+    if (attributes and any(a != "key" for a in attributes)) or event_handlers:
```
Archmonger: Is this change required? Using `any` here would be pretty expensive for large attribute dictionaries.

shawncrawley: The use of `any` certainly isn't required. You made a good point, so I've updated it accordingly to just do a `attributes.pop("key", None)` just before. A change of this nature is now necessary here due to changing the `pop` to a `get` in `Vdom.__call__`:
https://github.com/reactive-python/reactpy/blob/d457f61e6b8f39630eb45ff6389ed2e773009620/src/reactpy/core/vdom.py#L151

I wondered if I should compare the `key` kwarg with the potential `key` attribute in this `_fragment` method to ensure they are distinct - in case of a potential call like `_fragment({"key": "something"}, key="something-else")`, but then I realized that the call to this private `_fragment` method is already being controlled (as a private method should be), and in this case any potential `key` kwarg would always be coming directly from a `key` attribute provided to the `html.fragment` call (i.e. via that same `Vdom.__call__` line above). Thus, we just need to pop it out and throw it away here.

Let me know if this looks good and makes sense.

shawncrawley:
Very confused as to why a test just failed after my tweak... Is that some random fluke of a fail?

Archmonger (approved):

PR #1281, opened 2025-02-23
Rework VDOM construction and type hints

### LINKED ISSUE #1280: `reactpy.html` does not render nested lists
### Current Situation

The following snippet could be valid syntax, but will not render properly:

```python
html.div(
    [ "one", "two", ["three", "four"] ]
)
```

### Proposed Actions

Allow for nested lists to be recursively rendered as VDOM.

Archmonger:
## Description

- Removed ``reactpy.vdom``. Use ``reactpy.Vdom`` instead.
- Removed ``reactpy.core.make_vdom_constructor``. Use ``reactpy.Vdom`` instead.
- Removed ``reactpy.core.custom_vdom_constructor``. Use ``reactpy.Vdom`` instead.
- ``reactpy.html`` will now automatically flatten lists recursively (ex. ``reactpy.html(["child1", ["child2"]])``)
- Added type hints to ``reactpy.html`` attributes.

## Checklist


src/reactpy/core/vdom.py
```diff
@@ -124,148 +129,138 @@ def is_vdom(value: Any) -> bool:
...
             continue
 
-        separated_event_handlers[k] = handler
+        _event_handlers[k] = handler
+
+    return _attributes, _event_handlers
 
-    return separated_attributes, dict(separated_event_handlers.items())
+
+def _flatten_children(children: Sequence[Any]) -> list[Any]:
```
JamesHutchison, line 226: Could we use `itertools.chain.from_iterable` for the implementation instead? It's my understanding that it was written in C and is likely appreciably more performant

edit: except we don't want to split up strings. hmmm

JamesHutchison: Nevermind. The problem is more complicated than at first appearance and there doesn't appear to be a readily available solution to this problem. `from_iterable` is also not recursive

src/reactpy/core/vdom.py
```diff
@@ -124,148 +129,138 @@ def is_vdom(value: Any) -> bool:
...
+    def __init__(
+        self,
+        /,
+        allow_children: bool = True,
+        custom_constructor: CustomVdomConstructor | None = None,
+        **kwargs: Unpack[_VdomDict],
+    ) -> None:
+        """This init method is used to declare the VDOM dictionary default values, as well as configurable properties
+        related to the construction of VDOM dictionaries."""
+        if "tagName" not in kwargs:
```
JamesHutchison: could do `if not (self.__name__ := kwargs.get("tagName"):`?

JamesHutchison (commented):
I'm presuming the tagName change from first arg to kwarg is about streamlining the dict creation?

looks good

src/reactpy/core/vdom.py
Archmonger: > I'm presuming the tagName change from first arg to kwarg is about streamlining the dict creation?

Yes - Original intention was to make `Vdom` a subclass of `dict`, but I had reverted that. Now it's just a `dict`-like interface within `__init__`.

> could do `if not (self.__name__ := kwargs.get("tagName"):`?

Won't do this due to my personal opinion on bad readability on the walrus operator. Also, the current implementation is likely a few nanoseconds faster since it only needs to perform one attribute lookup, rather than two.

JamesHutchison: I can't argue on personal readability choices but I think it's an additional hash look-up to not use the walrus.

Another consideration is using slots to reduce memory usage and attribute look-up performance, although tbh I haven't kept up with the latest performance optimizations. I wouldn't normally nit pick performance but this is the core object of ReactPy that's used very frequently and repeatedly?

Archmonger: I'll spend some time to think about whether to add `__slots__`.

Under normal scenarios, the `Vdom` object itself isn't recreated frequently - It's initialized only once per `tagName` [when `reactpy.html` is called.](https://github.com/Archmonger/reactpy/blob/398dc2ae8ad69258bac668a562ce97cdb876636c/src/reactpy/_html.py#L292-L302)

The optimization would make a difference if we encourage/document how users can manually construct a `Vdom` element for performance reasons, and users start directly calling `Vdom( ... )` in their components.

JamesHutchison: oh yeah in that case if its once per tag then performance isn't a concern in the initialization. I wouldn't worry about it. Would calling it `VdomTagBuilder` (or several other potential names) be more appropriate? I guess the confusion for me originated from the change to use `**kwargs` rather than simply having the usual args that IDE's know how to handle and hint on. I misinterpreted it as being done for performance reasons.

src/reactpy/core/vdom.py
```diff
@@ -124,148 +129,138 @@ def is_vdom(value: Any) -> bool:
...
+                event_handlers=event_handlers,
+            )
+        # Otherwise, use the default constructor
+        else:
+            result = {
+                **({"key": key} if key is not None else {}),
+                **({"children": children} if children else {}),
+                **({"attributes": attributes} if attributes else {}),
+                **({"eventHandlers": event_handlers} if event_handlers else {}),
+            }
```
JamesHutchison: Consideration - if we're using a class then we could use inheritance:

```
CommonVdomBuilder -> ScriptVdomBuilder
                  -> FragmentVdomBuilder
```

tests/test_core/test_layout.py
```diff
@@ -82,7 +82,7 @@ async def test_simple_layout():
     @reactpy.component
     def SimpleComponent():
         tag, set_state_hook.current = reactpy.hooks.use_state("div")
-        return reactpy.vdom(tag)
+        return reactpy.Vdom(tagName=tag)()
```
JamesHutchison: What was the motivation to use `tagName` over `tag`? We appear to be violating pep8 with the camelCase. Since the `__init__` is a one time cost per tag, I would think we would want to stick to python snake case convention?

src/reactpy/core/vdom.py
Archmonger: `VdomDictConstructor` would be the appropriate name if we're following existing naming conventions within ReactPy. I didn't name it that since that's a bit wordy - I'll take some time to rethink the naming of this.

src/reactpy/core/vdom.py
Archmonger: I think this comment is on the wrong LOC?

tests/test_core/test_layout.py
Archmonger: I agree that it's strange and I'd like to change it.

It's a legacy interface created by Ryan, presumably designed from the perspective of the ReactJS client. On the client side, JSON objects traditionally follow `camelCase` for their keys.

Changing this would require modifying the VDOM spec, which would be a whole different PR's worth of work.

PR #1290, opened 2025-04-01
Support inline JavaScript events (v2)

shawncrawley:
## Description

This is an alternative solution to PR #1289, so only one of them need be merged. I created this separately to better compare approaches. This approach modifies the VDOM schema to include a new `jsExecutables` property alongside `eventHandlers`.

## Checklist


shawncrawley:
This is ready to be looked at - again, aside from the changelog.

I can get this to work "out of the box" with `string_to_reactpy` if I simply loosen the "on" event listener regex to not expect camelCase, which is usually how raw HTML will be written. The question is if we should coerce all properties starting with "on" to camelCase on the Python/Vdom side (e.g. "onclick" to "onClick"). Right now I'm not doing that, which is how I've written the `string_to_reactpy` test to expect the Vdom result. 

I ultimately assumed that coercing them to camelCase on the Vdom side is unnecessary overhead. Let me know if you agree.

Archmonger:
`onclick` should already be automatically mutated into `onClick` by `string_to_reactpy` at [this LOC](https://github.com/reactive-python/reactpy/blob/c890965d1c8189d7d8ef04cdf6554de80f0251b7/src/reactpy/transforms.py#L39-L48).

src/js/packages/@reactpy/client/src/vdom.tsx
```diff
@@ -198,23 +203,43 @@ function createEventHandler(
...
+      if (!(typeof value === "object" && value.nativeEvent)) {
+        return value;
+      }
+      const event = value as React.SyntheticEvent<any>;
+      if (preventDefault) {
+        event.preventDefault();
+      }
+      if (stopPropagation) {
+        event.stopPropagation();
+      }
```
Archmonger, line 218: What is your anticipated method for having users modify `preventDefault` and `stopPropagation` under this new system?

Would it not be better to have users handle it exclusively through JavaScript? For example `this.event.preventDefault()`?

src/reactpy/types.py
```diff
@@ -871,6 +880,7 @@ class VdomJson(TypedDict):
     children: NotRequired[list[Any]]
     attributes: NotRequired[VdomAttributes]
     eventHandlers: NotRequired[dict[str, JsonEventTarget]]
+    jsExecutables: NotRequired[dict[str, JavaScript]]
```
Archmonger: I'm not the biggest fan of the `jsExecutables` nomenclature, since it doesn't seem to make it "blindingly obvious" to the uneducated reader as to what this is.

Maybe `jsEventHandlers` is the more appropriate name? Then, we can rename the existing `eventHandlers` to `pythonEventHandlers`?

src/reactpy/types.py
```diff
@@ -885,6 +895,12 @@ class JsonImportSource(TypedDict):
     fallback: Any
 
 
+class JavaScript(str):
+    """A simple way of marking JavaScript code to be executed client-side"""
```
Archmonger:
```suggestion
    """Simple subclass that flags a user's string in ReactPy VDOM attributes as executable JavaScript."""
```

tests/test_utils.py
```diff
@@ -194,6 +194,15 @@ def test_string_to_reactpy(case):
                 "key": "my-key",
             },
         },
+        # 9: Includes `jsExecutables` attribue
+        {
+            "source": """<button onclick="this.innerText = 'CLICKED'">Click Me</button>""",
+            "model": {
+                "tagName": "button",
+                "jsExecutables": {"onclick": "this.innerText = 'CLICKED'"},
```
Archmonger: This really should have [been mutated](https://github.com/reactive-python/reactpy/blob/c890965d1c8189d7d8ef04cdf6554de80f0251b7/src/reactpy/transforms.py#L39-L48) within `string_to_reactpy` into `onClick`... Very odd that it didn't...

tests/test_core/test_events.py
```diff
@@ -221,3 +221,97 @@ def outer_click_is_not_triggered(event):
     await inner.click()
 
     await poll(lambda: clicked.current).until_is(True)
+
```
Archmonger, line 224: It would be good to add a test for a non-`^on[A-Z]` event handler too

src/reactpy/core/vdom.py
```diff
@@ -23,12 +24,16 @@
...
     ImportSourceDict,
+    JavaScript,
+    JSExecutableDict,
     VdomAttributes,
     VdomChildren,
     VdomDict,
     VdomJson,
 )
 
+EVENT_ATTRIBUTE_PATTERN = re.compile(r"^on\w+")
```
Archmonger: This can be reverted to `^on[A-Z]\w+` once we figure out why `string_to_reactpy` isn't properly mutating `onclick` to `onClick`.

Archmonger (changes requested):

src/js/packages/@reactpy/client/src/vdom.tsx
shawncrawley: I didn't add that, that's original code from the standard `eventHandler`. It shows as additions because I did pull it directly out of the return statement as an anonymous function and gave it a name - all so I could tack on a `isHandler = true` property (see line 222 below). This is how I chose to distinguish between the standard `eventHandler` and this new `jsExecutable` in the `react.js` template used for web components:
image

Alternatively, I had done `value.toString().includes('.sendMessage')`, but thought that was a bit expensive to do every time. Maybe not?

src/reactpy/types.py
shawncrawley: That was actually my exact initial leaning. I only hesitated because 1) I didn't want to necessarily change anything existing and 2) I suppose I don't fully like `eventHandler` for this javascript concept, again, because of my main use case where it's not being executed in response to an event at all, but simply as configuration. But I could be persuaded easily, since it was mainly my first reason that swayed me. If you're good with that, I can make the change. I even thought `pyEventHandlers` and `jsEventHandlers`, to match them up a bit more. Final thoughts?

tests/test_core/test_events.py
shawncrawley: I do, but I put it in `test_module.py` since I could only come up with a working example using a web module. It's called `test_callable_prop_with_javacript` - see below.

src/reactpy/types.py
Archmonger: I agree that `eventHandlers` isn't the right name either... Need to brainstorm some better names.

Perhaps `inlineJavascript`?

shawncrawley: Hm... Yeah, I like `inlineJavascript`. I can go ahead and make that change. If we come up with something better later, then I don't see changing it being a big deal for users. Most of the Vdom stuff is black-boxed to the user.

tests/test_utils.py
shawncrawley: You said this above:
> `onclick` should already be automatically mutated into `onClick` by `string_to_reactpy` at [this LOC](https://github.com/reactive-python/reactpy/blob/c890965d1c8189d7d8ef04cdf6554de80f0251b7/src/reactpy/transforms.py#L39-L48).

That LOC does not do the trick because that `RequiredTransforms` function is not executed until _after_ the Vdom constructor is called. And either way, that function only processes "attributes" not "jsExecutables". See these lines here:
https://github.com/reactive-python/reactpy/blob/c890965d1c8189d7d8ef04cdf6554de80f0251b7/src/reactpy/utils.py#L152-L156

Line 153 calls the Vdom constructor for the html element in question, and it's in the Vdom constructor that attributes are separated from handlers - including now these inline javascript cases. It's the `EVENT_ATTRIBUTE_PATTERN` (i.e. `r"^on[A-Z]+"`) that is used to decide which to separate out and cast as JavaScript. If I have the camelCase pattern, then "onclick" is not recognized, and so it remains in the "attributes" Vdom category, rather than being placed in "jsExecutables" where it belongs. 

It's a catch 22 in a way. It either ends up in "attributes" as "onClick" or "jsExecutables" as "onclick". We either have to convert to camelCase before calling the Vdom constructor to keep the camelCase regex check for identifying "jsExeutables" or we keep it loose and convert it to camelCase afterward - maybe by adding "jsExecutables" in addition to "attributes" in that LOC you referenced above.

Archmonger: A more healthy solution is to use `EVENT_ATTRIBUTE_PATTERN =  re.compile(r"^on[A-Z]\w+")`, then add a new function within `RequiredTransforms` that runs **after** `html_props_to_reactjs` that will perform the `EVENT_ATTRIBUTE_PATTERN` related transforms.

It's slightly less efficient to do it that way, but that's moot since `string_to_reactpy` is already wildly inefficient. Future me will likely rewrite `string_to_reactpy` in C, but today's goal is to maximize compatibility and limit future bugs.

src/reactpy/core/vdom.py
shawncrawley: Done.

tests/test_utils.py
shawncrawley: Done.

src/reactpy/types.py
shawncrawley: Done.

shawncrawley:
All comments have been addressed, changelog is updated... This PR should be ready to go.

src/js/packages/@reactpy/client/src/vdom.tsx
```diff
@@ -189,6 +189,12 @@ export function createAttributes(
           createEventHandler(client, name, handler),
         ),
       ),
+      ...Object.fromEntries(
+        Object.entries(model.inlineJavascript || {}).map(
+          ([name, inlineJavaScript]) =>
+            createInlineJavascript(name, inlineJavaScript),
```
Archmonger: Minor nitpick about capitalization consistency. Either choose `Javascript` or `JavaScript`. This same sentiment applies to every use of `javascript` within this PR.

src/reactpy/core/vdom.py
```diff
@@ -23,12 +24,16 @@
...
     ImportSourceDict,
+    InlineJavascriptDict,
+    JavaScript,
     VdomAttributes,
     VdomChildren,
     VdomDict,
     VdomJson,
 )
 
+EVENT_ATTRIBUTE_PATTERN = re.compile(r"^on[A-Z]\w+")
```
Archmonger, line 35: Rename to `INLINE_JAVASCRIPT_ATTRIBUTE_RE`

src/reactpy/transforms.py
```diff
@@ -37,14 +37,19 @@ def normalize_style_attributes(self, vdom: dict[str, Any]) -> None:
...
+        This method is private because it is called prior to instantiating a
+        Vdom class from a parsed html string, so it does not need to be called
+        as part of this class's instantiation (see comments in __init__ above).
+        """
+        attrs = cast(VdomAttributes, attributes.items())
+        attrs = cast(
+            VdomAttributes,
+            {REACT_PROP_SUBSTITUTIONS.get(k, k): v for k, v in attrs},
+        )
+        return attrs
```
Archmonger: Revert this function and create a separate transform function within `RequiredTransforms` that performs the `InlineJavascript` transforms.

The cadence set within `RequiredTransforms` is that it never needs any external calls, it just "does everything" automatically during initialization.

Archmonger (changes requested):
Can be merged after handling a few more nitpicks

src/js/packages/@reactpy/client/src/vdom.tsx
```diff
@@ -198,23 +204,43 @@ function createEventHandler(
...
+      }
+    }
+    if (args.length > 0 && args[0] instanceof Event) {
+      return handleExecution.call(args[0].currentTarget, ...args);
+    } else {
+      return handleExecution(...args);
+    }
+  };
+  wrappedExecutable.isHandler = false;
+  return [name, wrappedExecutable];
```
Archmonger: Since there's a couple of logical paths here, some comments would be appreciated on this function

src/reactpy/types.py
```diff
@@ -885,6 +895,12 @@ class JsonImportSource(TypedDict):
     fallback: Any
 
 
+class JavaScript(str):
+    """Simple subclass that flags a user's string in ReactPy VDOM attributes as executable JavaScript."""
+
+    pass
+
+
```
Archmonger: Should rename this to `InlineJavaScript` for consistency. Would also help users understand that `JavaScript` is not an "executor function", and as such the following scenario is **not** valid syntax:

```python
@component
def my_example():
    def on_click(event):
        result = JavaScript("this.innerText = 'hello'")
    ...
```

src/js/packages/@reactpy/client/src/vdom.tsx
shawncrawley: Good catch - I didn't notice I had been inconsistent myself. I'll go with JavaScript.

src/reactpy/transforms.py
shawncrawley: I'm not sure what to do here. I know about the RequiredTransforms cadence due to the comment in its `__init__` function, but to handle the "onclick" vs "onClick", "attributes" vs "inlineJavaScript", and `^on\w+` vs `^on[A-Z]\w+` dilemma (which we decided to stick with camelCase regex), I have to convert the inline html attributes to camelCase _prior_ to instantiating the Vdom object, which is also _prior_ to calling the RequiredTransforms - since these transforms all operate on Vdom object. Note the call order below:

image

I injected the camelCase transform to occur on the parsed HTML etree object (again, prior to it being used to instantiate the Vdom object and applying the standard RequiredTransforms. And if conversion to camelCase happens prior to Vdom instantiation, then there's no need to execute it again after. That is why I deleted the old function and just created a new "private" one, where it's only private so that it won't be invoked within the official cadence.

Does this makes sense? So again, I'm a bit confused. Do you want me to just restore the old function in addition to keeping the new one while maintaining the call order I've implemented? In which case it will perform a camelCase conversion twice - once before Vdom and once after.

shawncrawley: @Archmonger Just waiting on some clarification above.

Archmonger: Will respond on Friday. Been living and breathing my day job for the last few days.

Archmonger: Apologies for the late response, I finally managed to get some time away from work 😓.

To maintain the convention within `RequiredTransforms`, all we need to do here is hoist `attributes_to_reactjs` out of the class. It should become an independent function within the `transforms.py` module.

docs/source/about/changelog.rst
```diff
@@ -30,6 +30,7 @@ Unreleased
 - :pull:`1281` - Added ``reactpy.Vdom`` primitive interface for creating VDOM dictionaries.
 - :pull:`1281` - Added type hints to ``reactpy.html`` attributes.
 - :pull:`1285` - Added support for nested components in web modules
+- :pull:`1289` - Added support for inline JavaScript as event handlers or other attributes that expect a callable
```
Archmonger:
```suggestion
- :pull:`1289` - Added support for inline JavaScript as event handlers or other attributes that expect a callable via ``reactpy.types.InlineJavaScript``
```

Archmonger (approved):

